PR #252, opened 2026-05-05
fix(streeling-daily): null-directive filter + tetravalent → hexavalent

spareilleux:
## Summary

- Filters the `null → null` (priority: high) directive row from harvest output by reading both `id`/`target_repo` and `directive_id`/`to` schemas, skipping entries where either resolves empty
- Updates 3 references from "tetravalent" to "hexavalent" (Demerzel CLAUDE.md declares hexavalent T/P/U/D/F/C as canonical; tetravalent T/F/U/C is the deprecated 4-valued model)
- Switches the Governance-Research Connections link from `logic/tetravalent-logic.md` to `logic/hexavalent-logic.md` (both files still exist; hexavalent is the active one)

## Context

Surfaced during 2026-05-05 manual digest regeneration (org discussions [#250](https://github.com/orgs/GuitarAlchemist/discussions/250) + [#251](https://github.com/orgs/GuitarAlchemist/discussions/251)) which inherited both bugs from the 2026-05-01 weekly digest [#244](https://github.com/orgs/GuitarAlchemist/discussions/244).

The null-directive bug is real schema variation: `directive-tars-grammar-governance.json` is a `capability_request` type and uses `directive_id` + `to`, while the other six directives use `id` + `target_repo`. The fix is defensive in the workflow rather than touching the data files.

## Out of scope

- Persona count discrepancy (digest 17 vs local 18) is not a workflow bug — reflects an unpushed `autoresearch-driver.persona.yaml`. Will self-resolve when that lands.
- ~220 other `tetravalent` references across skills/policies/schemas — separate ecosystem-wide migration.

## Test plan

- [x] Read all seven `directive-*.json` files; jq fallback resolves every one cleanly
- [x] Workflow YAML structure unchanged (only inline shell + heredoc bodies modified)
- [ ] Verify on next workflow_dispatch — the resulting digest should have **7 directives + zero `null` rows** (the fallback rescues the capability_request directive instead of dropping it)


spareilleux (commented):
Review failed: no response

---

PR #346, opened 2026-06-20
docs(context): seed CONTEXT.md domain glossary

spareilleux:
Minimal grounded domain glossary (the AX lever the engineering skills read). Grow lazily.


spareilleux:
Closing — the branch is stale vs master (it would revert ~52 unrelated trigger files). master already has a CONTEXT.md stub, so the AX goal is met; a grounded-glossary upgrade can be done cleanly as a 1-file change later.

PR #363, opened 2026-06-21
fix(halt): jsonschema-with-stdlib-fallback validation parity (ADR-0003)

spareilleux:
Candidate 4 (halt half) of the 2026-06-21 architecture review. The digest half shipped with Candidate 1 (DigestState).

`demerzel_halt.py` harvested the enum/length from the schema but its hand-rolled `validate()` enforced a **looser subset** than the PowerShell `Test-Json` adapter (no `pattern`, `additionalProperties`, or `required`-list check) — so the two adapters could diverge.

- `validate()` now uses `jsonschema` when importable (full parity with `Test-Json`), else a stdlib generic schema-subset check. The fallback is kept because the halt tool is the cross-repo **emergency brake** and must run on bare Python without `pip install`.
- `scripts/requirements-halt.txt` declares `jsonschema` (pinned) so CI takes the strong path.
- `tests/python/test_demerzel_halt_validate.py` proves both paths agree, incl. new parity catches (malformed timestamp, additional field).
- ADR-0003 amended (refines, does not reverse the stdlib-only stance).

**Verified:** 7/7 validate tests pass under both paths; CLI `status` works.

### DIFF
--- a/docs/adr/0003-schema-single-source-validation.md
+++ b/docs/adr/0003-schema-single-source-validation.md
@@ -61,3 +61,17 @@ follow-on), not by the schema.
 - **Keep per-language hand-coded validators, accept the duplication.** Rejected: the halt
   marker is the proof the duplication drifts; the schema is readable by both runtimes, so the
   shared seam costs little.
+
+## Update (2026-06-21)
+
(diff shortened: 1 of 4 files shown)


spareilleux (commented):
Review failed: no response

---

PR #373, opened 2026-06-21
feat(ml-feedback): bounded self-improvement loop — calibration + staleness + violation-pattern (3 legs live)

spareilleux:
Turns two legs of Demerzel's ML-governance feedback loop from 🔴 declared-only into 🟢 live, on **real data**. Producer half lives in ix (companion PR GuitarAlchemist/ix#168); this is the **governor** half + the wiring.

## What's now live
- **§3a confidence-calibration** — ix computes overconfidence from belief revision history → Demerzel applies a bounded threshold nudge (±0.1 cap).
- **§3b staleness** — ix predicts stale beliefs → Demerzel schedules capped proactive recons (3/day).

## Governor: `scripts/apply_ml_feedback.py`
Executes the governance steps `pipelines/ml-feedback-loop.ixql` §2–§6 describes but that no IxQL runtime exists to run (it stands in for that absent executor on these legs). Modeled on `scripts/demerzel_halt.py`.
- **Integrity gate** — recomputes & verifies `content_hash` (anti-tamper).
- **Article-9 safety gate** — drops `modify_constitution`/`override_policy`, requires confidence ≥0.7 and `constitutional_check.passed`; governance gate ≥0.8; bounded auto-apply ≥0.85, else escalate to `pending-human-review`.
- **Type dispatch** — `threshold_adjustment` → bounded belief nudge; `proactive_recon` → capped recon schedule; any type with no applier escalates rather than guessing.
- **Reversibility (Art.3)** prior values recorded; **transparency (Art.2)** every applied update carries its rationale; **auditability (Art.7)** appended to `state/evolution/`.

## Also: real schema/pipeline drift fix
`ml-feedback-loop.ixql` §3a–§3d filtered `recommendation_type` on the *policy's* output-type names (`calibration_report`, `staleness_forecast`, `pattern_analysis`, `strategy_report`) — **none** are in the validated schema enum, so a schema-valid doc never matched. Corrected each to its schema-valid value.

## Verification
- End-to-end on real data: 16 beliefs → overconfidence 0.5 → −0.1 nudge applied; 13 beliefs → 12 months-stale → top-3 recons scheduled.
- Negative-tested: tampered hash / low confidence / forbidden type all blocked.
- 3/day recon cap holds across repeated cycles (re-run schedules nothing new).

## Scope / honesty
[...]

PR #375, opened 2026-06-21
feat(ml-feedback): branch-independent orchestrator + finish audit-log move

spareilleux:
Follow-up to #373 with two fixes.

## 1. Orchestrator hardening (the requested caveat)
`run_ml_feedback_cycle.py` now sources the four ix producers from ix **`origin/main`** (materialized into `state/.cache/ix-producers/`) rather than ix's working tree. A scheduled cycle is now independent of whichever branch ix is mid-feature on, and won't pick up uncommitted producer edits. Falls back to the working tree if git/origin-main is unavailable (offline). New `--producer-source {main,worktree}` (default `main`).

**Verified:** with **zero** working-tree producers present, the cycle still ran all four from `origin/main` → applied=3, escalated=1, 0 errors.

## 2. Completes the audit-log move #373 only half-landed
In #373, an aborted `git add` (a `fatal: pathspec` on the already-deleted evolution file) silently dropped two staged changes, so master still has `_append_evolution` writing to `state/evolution/` — which would **recreate the schema-violating file** the governance validator rejects. This commits the dropped pieces: the `_append_audit` rename (→ `state/oversight/ml-feedback-audit.json`) and the `.gitignore` entries (`ml-feedback-audit.json`, `state/.cache/`).

`validate_governance.py` passes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,5 @@ state/oversight/scheduled-recons-*.json
 state/oversight/compliance-reports/*.json
 state/oversight/policy-review-requests.json
 state/oversight/ml-feedback-cycle-*.json
+state/oversight/ml-feedback-audit.json
+state/.cache/
--- a/scripts/apply_ml_feedback.py
+++ b/scripts/apply_ml_feedback.py
@@ -310,8 +310,12 @@ _APPLIERS = {
 }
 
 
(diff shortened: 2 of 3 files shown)


spareilleux (commented):
Review failed: no response

---

Issue #381, opened 2026-06-23
AFK smoke: add a one-line 'first proven' note to docs/agents/afk-agent.md


spareilleux:
Append a single line at the end of the Contract section in docs/agents/afk-agent.md noting the AFK agent was first proven end-to-end on 2026-06-22. Trivial, low-risk documentation change only.

spareilleux:
🤖 **Demerzel Governance Notice**

This issue has been added to the [Demerzel Governance](https://github.com/orgs/GuitarAlchemist/projects/2) project board.

**Governance checklist:**
- [ ] Reviewed against [Asimov constitution](https://github.com/GuitarAlchemist/Demerzel/blob/master/constitutions/asimov.constitution.md) (Articles 0-5)
- [ ] Checked [default constitution](https://github.com/GuitarAlchemist/Demerzel/blob/master/constitutions/default.constitution.md) compliance (Articles 1-11)
- [ ] Risk assessed per [autonomous loop policy](https://github.com/GuitarAlchemist/Demerzel/blob/master/policies/autonomous-loop-policy.yaml)
- [ ] [Reconnaissance](https://github.com/GuitarAlchemist/Demerzel/blob/master/policies/reconnaissance-policy.yaml) completed if cross-repo

*Governed by Demerzel under the [governance mandate](https://github.com/GuitarAlchemist/Demerzel/blob/master/constitutions/demerzel-mandate.md).*

PR #380, opened 2026-06-23
feat(afk): AFK agent (queue→sandbox→PR) + Pocock harness doctrine

spareilleux:
## What & why

Absorbs the Matt Pocock × David Ondrej agentic-engineering talk
(`sources/chats/matt-pocock-david-ondrej-agentic-workflow.md`) and turns it into
two deliverables. Design + plan: `docs/superpowers/specs/2026-06-22-…` and
`docs/superpowers/plans/2026-06-22-…`.

### 1. Doctrine reconciliation (Deliverable 1)
New **"Harness doctrine (Pocock delta)"** section in `CLAUDE.md` reconciling
Matt's principles with the existing Karpathy/Cherny doctrine (agree=keep,
diverge=adjust): harness≈model & agent-agnostic, **queue-not-loop**,
**procedures-over-abilities** (context-leak cost), delete→observe→layer-back,
DX≈AX. Plus a `/learnings` capture.

### 2. AFK agent (Deliverable 2)
Closes the one missing link in our existing autonomous machinery — the
**implement** step between the trigger queue (`demerzel-driver-triggers`) and the
PR review gates (`agent-blackbox`, `cross-model-review`):

- **`scripts/run_afk_cycle.py`** — stdlib governor (mirrors the proven
  `run_ml_feedback_cycle.py`): HALT-honored, risk classification
  (constitution/policy = critical = never auto-implemented), Podman preflight,
  harness invocation, PR creation, loop-state + `state/oversight/` audit. 9/9 unit tests.
- **`prompts/afk-implement.prompt.md`** — the sandboxed agent's operating
  instructions (scope/simplicity guardrails + oracle gate on `validate_governance.py`).
- **`docs/agents/afk-agent.md`** + `agent-implement` label in `triage-labels.md`.

### Sandbox backend: Podman (not Docker)
Docker Desktop is fragile across reboots; we use **Podman** (daemonless, no
Desktop app) with a `_ensure_podman()` preflight that starts the machine
idempotently.

## Proof
The riskiest unknown is **solved**: the sandcastle+Podman harness ran headless
Claude Code in a container on Windows and produced a real commit
(Windows↔WSL2 path translation works). Final whole-branch review (opus) =
merge-ready after 4 fixes (all applied: no-op/pr-fail status correctness +
Docker→Podman doc accuracy).

## Notes for reviewers
- The **runtime harness lives OUTSIDE this repo** at `../afk-harness/`
[...]

PR #384, opened 2026-06-23
fix(afk): invoke harness via node --import tsx, not npx (Windows)

spareilleux:
Follow-up to #380. The end-to-end dogfood (issue #381) surfaced a Windows-only bug in the governor: `_invoke_harness` ran `["npx", "tsx", ...]` via Python `subprocess` (no shell). On Windows `npx` is `npx.cmd` (no `.exe`), so `CreateProcess` fails with WinError 2 — and using a shell would expose the issue body to command injection.

Fix: invoke `node --import tsx <main.mts>` directly. `node.exe` is a real executable and all args pass through as argv (no shell, no injection surface).

**Proven:** after this fix the AFK agent ran the full path autonomously — picked #381 off the queue, implemented it in the Podman sandbox, and opened #383.

### DIFF
--- a/scripts/run_afk_cycle.py
+++ b/scripts/run_afk_cycle.py
@@ -108,8 +108,14 @@ def build_loop_state(issue: dict, seq: int, risk: str, mode: str, today: str) ->
 
 
 def _invoke_harness(issue: dict) -> dict:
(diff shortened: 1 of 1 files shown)


spareilleux (commented):
Review failed: no response

---

PR #648, opened 2026-07-02
[AIW-v0.1][P0] Prompt and harness engineering discipline for AFK agents

spareilleux:
Adds prompt pack logic (`scripts/aiw_prompt_pack.py`) and result schema (`schemas/aiw-harness-result.schema.json`) for AIW execution episodes. Updates `README.md` and generates the manifest. Includes tests enforcing proper execution.

---
*PR created automatically by Jules for task [13346103974311787099](https://jules.google.com/task/13346103974311787099) started by @spareilleux*

Issue #461, opened 2026-06-27
[AIW-v0.1][P0] Prompt and harness engineering discipline for AFK agents


spareilleux:
Parent: #457
Related: #455, #459
Reference video from user: https://youtu.be/nQwJVHCtDDY?si=3RoGRwmUx9vpBfYF

## Goal

Improve Demerzel's prompt engineering and harness engineering so AIW workers become more reliable, testable, auditable, and AFK-capable.

The goal is not to create longer prompts. The goal is to create better task contracts, better harnesses, better feedback loops, and smaller autonomous steps.

## Why

A multi-provider AI workforce will fail if each provider receives vague prompts and unbounded autonomy.

Demerzel needs a disciplined prompt/harness layer that turns GitHub issues into bounded executable jobs with:

- precise task framing;
- explicit context bundles;
- allowed paths;
- non-goals;
- test commands;
- expected outputs;
- failure modes;
- retry rules;
- budget limits;
- review evidence;
- AFK eligibility.

This should align with the Matt Pocock-style principles already captured in #455: small composable skills, shared language, issue shaping, TDD, bug diagnosis, codebase design, and avoiding vibe-coded broad changes.

## Core principle

A prompt is not enough. Every serious AIW task needs a harness.

A good harness makes the desired behavior observable:

1. The agent gets a crisp job spec.
2. The agent runs in a bounded workspace.
3. The agent must produce a structured result.
4. The result is validated by tests/checks/static review.
5. Failures are summarized and minimized before retry.
6. Escalation happens only when evidence justifies it.

## Prompt engineering requirements

Each AIW provider prompt should be generated from a reusable prompt pack, not hand-written ad hoc.

Minimum prompt sections:

```markdown
# Role
You are acting as: <provider-role>

# Task
<one bounded task>

# Source of truth
<link or embedded issue spec>

# Context bundle
<files, summaries, constraints, prior attempts>

# Allowed scope
<paths and operations allowed>

# Non-goals
<explicitly out of scope>

# Constraints
<budget, runner, sandbox, secrets, network, coding style>

# Required process
<small steps, tests first when applicable, no broad rewrites>

# Required outputs
<branch/patch/summary/tests/risk notes/ledger>

# Stop conditions
<missing context, failing tests, budget exceeded, risk escalation>
```

## Harness engineering requirements

Add a harness contract around every AFK-capable provider adapter.

The harness should provide:

- input materialization: issue body, source bundle, files, prompt pack;
[...]

PR #735, opened 2026-07-18
fix(aiw-prompt-pack): AIW prompt-pack + harness rules, CI-green (corrected #648)

spareilleux:
Corrected rebuild of #648 onto current master. Original failed because its test was **pytest**-style but the repo runs `unittest discover` (pytest not installed), and the branch was stale/conflicting.

- test converted to unittest (9 tests, no pytest/regex)
- `aiw_prompt_pack.py` schema path resolved from repo root (was cwd-fragile)
- carries only the real artifacts (module + test + `schemas/aiw-harness-result.schema.json`); manifest regenerated; README schema count bumped

**Evidence:** full suite **173 passed**; example validates against schema; governance + manifest green.

Supersedes #648. Refs #461

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ Demerzel/
 | Personas | 17 | `personas/*.persona.yaml` |
 | Policies | 46 | `policies/*.yaml` |
 | Grammars | 20 | `grammars/*.ebnf` |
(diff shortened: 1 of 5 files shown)


spareilleux (commented):
Review failed: no response

---

PR #758, opened 2026-07-19
feat(ci): scheduled-run health guard (green != dead for scanners)

spareilleux:
## Why
`demerzel-capability-expansion` was RED for ~a month (every scheduled run failed at exit 5) before anyone noticed — the "silent loop death" failure mode. The existing `demerzel-quality-trend-freshness` guard catches dead **data** feeders by file staleness, but **scanners** that only emit issues/discussions produce no file that goes stale, so their run failures were invisible.

This is the metafix for that: watch scheduled-run **health**, not just data freshness.

## What
A weekly, independent guard (`demerzel-scheduled-run-health.yml`) that:
- reads the latest **scheduled** run `conclusion` of every cron workflow;
- **silent on green**; turns **RED** and opens/updates one `loop-health` tracking issue when any last failed (`failure` / `timed_out` / `startup_failure`);
- posts a **Discord** alert too — **dormant until a `DISCORD_WEBHOOK_URL` repo secret is added** (logs a notice + no-ops until then);
- skips itself (no self-referential flapping); **never remediates** — a human decides the fix.
- Uses the default `GITHUB_TOKEN` (`actions:read` + `issues:write`), no PAT.

## Verification (against live run data, read-only)
Checked **12** scheduled workflows; stayed silent on the 10 green; correctly flagged:
- `demerzel-capability-expansion` — last scheduled run failed 07-13 (its fixed run lands next Monday)
- `ecosystem-freshness` — **a second silently-failing loop this guard surfaced on first run**

YAML validated. The `ecosystem-freshness` red is a pre-existing, separate issue worth its own triage — surfaced here as proof the guard works.

## Notes
- Aligns with the existing `demerzel-quality-trend-freshness` "green != dead" doctrine, generalised from data-staleness to run-health.
- To enable Discord: add repo secret `DISCORD_WEBHOOK_URL`.

### DIFF
--- a/.github/workflows/demerzel-scheduled-run-health.yml
+++ b/.github/workflows/demerzel-scheduled-run-health.yml
@@ -1,0 +1,117 @@
+name: Demerzel Scheduled-Run Health
+
+# Green != dead guard for SCHEDULED workflows.
+#
+# A scanner that fails silently is indistinguishable from a healthy one until you
+# look — that is exactly how `demerzel-capability-expansion` stayed RED for ~a
+# month in 2026 unnoticed (every scheduled run failed at exit 5, nobody saw it).
+# The existing `demerzel-quality-trend-freshness` guard catches dead DATA feeders
+# by file staleness, but scanners that only emit issues/discussions produce no file
+# that goes stale, so nothing watched their run health.
+#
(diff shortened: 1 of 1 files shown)


spareilleux (commented):
Review failed: no response

---

PR #759, opened 2026-07-19
Revert #758: scheduled-run health guard duplicates ecosystem-freshness

spareilleux:
## Revert rationale
PR #758 added `demerzel-scheduled-run-health.yml`, but that duplicates the existing **`ecosystem-freshness.yml`** guard (`scripts/ecosystem_freshness.py` — "Universal ecosystem-freshness monitor, one guard for every scheduled loop").

The existing guard is strictly more capable:
- **Daily** (vs #758's weekly);
- detects `failed_run` (latest scheduled run conclusion) **plus** staleness, `expired_disable`, `activation_mismatch`;
- backed by a **tested** script (`test_ecosystem_freshness.py`);
- **least-privilege two-job split** (read-only evaluate / write-only persist);
- persists a durable `state/loop-health/.freshness-alert.json` and self-clears on recovery.

#758 was a less-robust reimplementation shipped because the existing guard wasn't found before building. Reverting to avoid a permanent duplicate (LOLLI/duplication).

## Note
`ecosystem-freshness` is currently (correctly) RED because `demerzel-capability-expansion` and `demerzel-self-improvement` had failed scheduled runs — both now fixed; it will self-clear on its next daily run. If Discord alerting is wanted, the right place is a small addition to the existing guard's alert path, not a parallel workflow.


spareilleux (commented):
Review failed: no response

---

PR #760, opened 2026-07-19
feat(loop-health): Discord alert when ecosystem-freshness trips

spareilleux:
## What
Adds an optional **Discord** notification to the **existing** `ecosystem-freshness` guard — *not* a parallel workflow (that mistake was #758, reverted in #759).

When the `persist` job commits a new/changed trip, it emits an `alerted` output and a follow-up step posts the tripped workflows + kinds to Discord.

## Design
- **Fires only on a committed change** — the recovery/no-change paths `exit 0` before it, so no daily spam; you get pinged when a loop newly trips (or the set of red loops changes).
- **Dormant until `DISCORD_WEBHOOK_URL` secret is set** — logs a `::notice::` and no-ops otherwise, so this is a safe no-op until you add the secret.
- Message built from the **verified** committed alert schema (`state/loop-health/.freshness-alert.json` → `.problems[].workflow` + `.kind`), e.g.:
  ```
  🔴 Demerzel loop-health tripped
  - `demerzel-capability-expansion.yml` — failed_run
  ```

## Verification
YAML parses; the summary jq renders correctly against the real alert artifact. +24 lines, single file.

## To enable
Add repo secret **`DISCORD_WEBHOOK_URL`** (a Discord channel webhook URL).

### DIFF
--- a/.github/workflows/ecosystem-freshness.yml
+++ b/.github/workflows/ecosystem-freshness.yml
@@ -90,6 +90,7 @@ jobs:
           path: ${{ runner.temp }}/ecosystem-freshness-result
 
       - name: Persist failure or verified recovery
+        id: persist
         if: always()
         run: |
           result_dir="$RUNNER_TEMP/ecosystem-freshness-result"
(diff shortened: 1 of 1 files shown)


spareilleux (commented):
Review failed: no response

---

PR #772, opened 2026-07-20
feat(aiw): provider-native token accounting + knowledge-type routing doctrine

spareilleux:
Follow-up to #771 — executes two items from that PR's adoption table (P5, P6), grounded in Pocock's token + cost videos.

## P5 — provider-native token accounting (code)

Budget caps (`max_total_tokens`) are expressed in **canonical** tokens, but every provider ships its own tokenizer, so the same prompt bills a different count on each (**a 2–3× spread is ordinary**). A single normalized cap silently **over-admits** work on coarse tokenizers and **under-admits** it on fine ones — a latent correctness gap, most consequential for the metered providers.

Adds an optional per-provider `token_multiplier` to `state/driver/aiw-budget-policy.json` that converts a canonical estimate into the provider's own tokens before the cap check.

- **Absent means `1.0`** → every currently shipped provider keeps *exactly* its previous behaviour. Non-breaking by construction.
- Must be **positive and finite**; `0`, negatives, NaN, booleans and strings are rejected as invalid policy.
- Gate now reports `estimated_total_tokens` (raw) alongside `token_multiplier` and `effective_total_tokens`; `token_cap_exceeded` is raised against the **effective** count.

**Real multipliers are deliberately not guessed.** Populating them needs *measured* tokenizer ratios per provider; the doc asks for the measurement as evidence. Shipping the mechanism without inventing the data.

## P6 — knowledge-type routing (doctrine)

`cost-aware-routing.md` routes by **risk/stakes**, which tends to send every *hard* task to a premium model. Adds an orthogonal axis:

- **Parametric** knowledge (in the weights) → planning, grilling, architecture. A bigger model is genuinely better; you're paying for what it knows that you didn't put in the window.
- **Contextual** knowledge (in the prompt) → implementation. Once the plan and seams are decided, a cheaper model performs close to a frontier one.

**Rule:** plan with the strongest model, implement with the cheapest that clears the bar. A hard feature doesn't justify a premium *implementation* model — an under-specified one does; the first remedy for poor output is a better plan in context, not a bigger model.

Documented doctrine only. `aiw-worker-lanes.json`'s existing `work_class` field is noted as the natural hook to make it executable later — **the hardened lane selector is untouched**.

## Deliberately NOT built

The **monthly-reset AFK ledger** (also in the adoption table). Anthropic's AFK credit is monthly/non-rolling, but **no consumer tracks a monthly boundary today** — building it now would be speculative. Deferred as a proposal rather than shipped blind.

## Verification

- Budget gate **32 → 37** tests (5 new, written red-first — including a regression guard that the absent-field path is unchanged).
- **62** tests green across the full AIW suite (gate, consumer, lane selector, prompt pack) — no regressions.
- `validate_governance` 13 valid / 0 invalid; `build_manifest` green, no drift (regenerated `governance-health.json` reverted to keep the diff surgical).

### DIFF
--- a/docs/architecture/cost-aware-routing.md
+++ b/docs/architecture/cost-aware-routing.md
@@ -22,6 +22,39 @@ To avoid "LLM-by-default" patterns, the following tasks *must* attempt a determi
 - **Syntax Validation**: Use compilers, linters, or tree-sitter.
 - **Dependency Analysis**: Use repo-specific package managers or graph tools.
 
+## Routing by Knowledge Type (parametric vs contextual)
+
+The hierarchy above routes by **risk and stakes**. That is necessary but not
+sufficient: it tends to send *every* hard task to a premium model, because "hard"
+reads as "high-stakes". A second, orthogonal question sharpens it — **which kind of
+knowledge does this task actually draw on?**
+
+- **Parametric knowledge** lives in the model's weights (breadth, prior art,
+  unprompted alternatives, "have you considered X?"). **Planning, grilling,
+  architecture, and design exploration depend on it** — a bigger model is
+  genuinely better here, because you are paying for what it knows that you did not
+  put in the window.
+- **Contextual knowledge** lives in the prompt (the agreed plan, the spec, the
+  files already in context). **Implementation depends mostly on this** — once a
+  task is well-scoped and the plan and seams are decided, a cheaper model performs
+  close to a frontier one, because the hard thinking is already in the window.
+
+**The rule:** *plan with the strongest model available; implement with the cheapest
+model that clears the bar.* A difficult feature does **not** by itself justify a
+premium implementation model — an under-specified one does. If implementation
+quality is poor, the first remedy is a better plan/spec in context, not a bigger
+model.
+
+This composes with, and does not replace, the tier hierarchy: deterministic tools
+still come first, and a high-**risk** change still escalates for review regardless
+of which model produced the diff.
+
+> Wiring note: `state/driver/aiw-worker-lanes.json` already classifies every role by
+> `work_class` (`groom`/`review`/`navigate` are planning-shaped; `net_new_feature`/
+> `fix` are implementation-shaped). That field is the natural hook for making this
+> rule executable in `scripts/aiw_lane_selector.py`; it is documented doctrine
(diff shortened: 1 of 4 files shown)


spareilleux (commented):
Review failed: no response

---

spareilleux:
Closing this rather than patching it. Adversarial review found six defects; two were blocking, but the one that decides it is finding 1.

**The mechanism is attached to the dimension that costs nothing.** `token_multiplier` scales exactly one comparison — the token cap. It never touches `estimated_cost`, `max_cost`, cycle spend, the approval threshold, or `release()`. For a metered provider the harm of a coarser tokenizer *is* cost, and cost is supplied independently by the caller, so this cannot fix metered budgeting. The providers where a pure token cap binds are the `subscription-or-local` ones, which have no marginal spend. The PR's central justification was wrong.

**And it has no consumer.** Measured against the shipped policy:

- No provider declares `token_multiplier` (all 7 `ABSENT`)
- No provider overrides `max_total_tokens` — the `200000` cap is a single global default

So the feature is inert, while its live effects are both negative:

- **Unbounded cap-widening.** No schema exists for `aiw-budget-policy.json` at all. `multiplier 0.001` + 199,000,000 tokens returns `allow` under a 200k cap. Patching `0.001` onto the metered `jules` provider leaves all 41 tests green.
- **Fail-closed downgrade.** `int(effective_tokens)` can raise `OverflowError`, which is absent from `main()`'s except tuple, so exit 2 (invalid policy, fail closed) degrades to exit 1 (governed block, run stays green) — and the workflow then reads a stale ledger.

Shipping an unconsumed mechanism that widens a money gate to fix a problem it structurally cannot fix is not a trade worth making. Withdrawn.

The doc changes go with it: `aiw-budget-router.md` actively solicited these values into an agent-writable file, and the `cost-aware-routing.md` section contradicted the tier hierarchy three sections above it (`review` work_class is defined as "risk classification, summarization", which tier 2 routes to a *local* model).

Two real findings survive as their own work:

- No JSON Schema for `aiw-budget-policy.json` — the policy gating real spend is unvalidated. Filed as #784.
- If provider-native accounting is ever wanted, it must apply to **cost**, with a bounded multiplier and a schema, and be introduced with an actual consumer.

PR #775, opened 2026-07-20
refactor(docs): decompose CLAUDE.md into single-sourced authorities

spareilleux:
## What this is

`CLAUDE.md` 124 → 69 lines, plus five `.claude/rules/*.md` retired into the docs that already own their subject. The resident instruction surface duplicated several rules 2–4× — including two the repo had already *declared* single-sourced (`logic/confidence-thresholds.yaml`, the manifest-enforced persona→test rule). This applies Demerzel's own low-duplication principle to its own instruction surface.

## Corrected after adversarial review

The first version of this PR made two false claims. Both are fixed, and both are worth recording.

**"A hook now enforces the deterministic rules" — it did not.** The `PreToolUse` guard has been **removed**. Executed, it allowed `git push -u origin master`, `git push -f origin master`, and `git push origin HEAD` while on master — twelve bypass variants. Its runtime-code rule listed languages this repo doesn't use while omitting `.ps1`/`.psm1`/`.sh`, which is *every* executable file in this harness. It also blocked legitimate work: any `master-*` branch, and any doc quoting a push command.

This mattered beyond the hook itself: **no-runtime-code had been demoted from a resident rule on the grounds the hook enforced it.** Prose was traded for enforcement that didn't exist — worse than either alone. The prose is restored, and `CLAUDE.md` now states plainly that these prohibitions are *not* hook-enforced. Gap tracked in #786, which recommends the existing `git-guardrails-claude-code` skill over a bespoke regex.

Also dropped with it: the hook introduced a **new** prohibition on pushing to the default branch, justified by citing "the push half of `commit-style.md`" — which has no push half, and no such rule exists anywhere in the repo. That deserves its own decision, not a footnote in a de-duplication PR.

**"No information was deleted" — five facts landed nowhere.** All restored:

| Fact | Now lives in |
|---|---|
| Tracer-bullets + vertical slices | `docs/methodology/harness-driven-development.md` |
| Pocock harness doctrine (6 bullets) | same file — which `CLAUDE.md` already pointed at but did not contain |
| `/goal`, declared-vs-operational | `CLAUDE.md` Karpathy R4 |
| Policy versioning with rationale | `CONTRIBUTING.md` |
| `/learnings` output path | `CLAUDE.md`, now noting no such skill exists |

Tracer-bullets was the worst: `grep "Tracer-bullet first"` returned **zero hits repo-wide** for a rule the old file called non-negotiable.

Also fixed: three dangling `.claude/rules/*` citations in behavioral tests and a plan doc, and the `CONTEXT.md` / `CONTRIBUTING.md` back-pointers that still described `CLAUDE.md` as the full reference after it became an index — a circular pointer, in a PR about single-sourcing.

## Verified

- CI anchor `Karpathy 4 Rules` present (`karpathy-cherny-discipline.yml:44` greps for it)
- `## Session-learned rules` heading exact — `/correct` appends under it
- `.claude/settings.json` now byte-identical with master
- No remaining dangling `.claude/rules/*` references
- Other `CLAUDE.md` consumers read *sibling repos'* files, unaffected


spareilleux (commented):
Review failed: no response

---

Issue #782, opened 2026-07-20
Overseer reports oracle green from a two-month-stale artifact


spareilleux:
## Summary

`state/quality/demerzel-harness/last.json` is a tracked, hand-written artifact that asserts the oracle is passing. Nothing recomputes it. The overseer reads it and reports green while the oracle is red.

## Measured

```json
{
  "domain": "demerzel-harness",
  "emitted_at": "2026-05-16T22:20:00Z",
  "metric_name": "harness_ready",
  "metric_value": 1.0,
  "oracle_status": "ok",
  "oracle_command": "pwsh scripts/verify.ps1",
  "summary": "Harness baseline established for Agent Blackbox readiness."
}
```

Last modified 2026-05-17 (`10de265`). The oracle has been red since **2026-06-21** (`58ff08c`, #371, which added the lockfile that `ConvertFrom-Json` rejected). `dev-process-overseer.ps1` surfaces `oracleStatus: ok` sourced entirely from this file.

So for roughly two months a green light has been reported by an artifact that predates the breakage and is never refreshed.

## Why this is the interesting failure

This is the exact shape the repo already has doctrine for: a state producer with no independent freshness guard, where **green ≠ alive**. The oracle's actual status was recoverable at any time by running it — nothing did, and nothing checked that the recorded status had an expiry.

Note the layering: even once #778 lands and the oracle reports honestly, *this* file will still say `ok`, because it is not derived from an oracle run at all.

## Suggested fix

Two independent changes, both needed:

1. **Derive it.** `oracle_status` must come from an actual `verify.ps1` invocation, not a committed literal. Either emit it as part of the run, or stop recording it and have the overseer invoke the oracle.
2. **Expire it.** Give the overseer a staleness bound on `emitted_at` and treat an over-age artifact as `unknown`, never `ok`. Fail toward "I don't know," not toward green.

## Related

- #778 — the oracle fix that exposed the divergence
- #780 — the grammar defect the oracle was hiding

spareilleux:
🤖 **Demerzel Governance Notice**

This issue has been added to the [Demerzel Governance](https://github.com/orgs/GuitarAlchemist/projects/2) project board.

**Governance checklist:**
- [ ] Reviewed against [Asimov constitution](https://github.com/GuitarAlchemist/Demerzel/blob/master/constitutions/asimov.constitution.md) (Articles 0-5)
[...]

Issue #703, opened 2026-07-18
triage(loop): revive capability-expansion after four consecutive scheduled failures


spareilleux:
## Evidence

All four available scheduled `demerzel-capability-expansion.yml` runs failed, most recently 2026-07-13: https://github.com/GuitarAlchemist/Demerzel/actions/runs/29270858003

## Recommended disposition: revive

Keep the capability-expansion outcome, but fail visibly until the scheduled producer demonstrates a successful run.

## Acceptance

- [ ] Diagnose and fix the minimum causal failure.
- [ ] Cover the boundary with an offline test or validation fixture.
- [ ] A manual canary and the next scheduled run succeed.
- [ ] No new provider secret; use only existing Demerzel-scoped credentials.

spareilleux:
🤖 **Demerzel Governance Notice**

This issue has been added to the [Demerzel Governance](https://github.com/orgs/GuitarAlchemist/projects/2) project board.

**Governance checklist:**
- [ ] Reviewed against [Asimov constitution](https://github.com/GuitarAlchemist/Demerzel/blob/master/constitutions/asimov.constitution.md) (Articles 0-5)
- [ ] Checked [default constitution](https://github.com/GuitarAlchemist/Demerzel/blob/master/constitutions/default.constitution.md) compliance (Articles 1-11)
- [ ] Risk assessed per [autonomous loop policy](https://github.com/GuitarAlchemist/Demerzel/blob/master/policies/autonomous-loop-policy.yaml)
- [ ] [Reconnaissance](https://github.com/GuitarAlchemist/Demerzel/blob/master/policies/reconnaissance-policy.yaml) completed if cross-repo

*Governed by Demerzel under the [governance mandate](https://github.com/GuitarAlchemist/Demerzel/blob/master/constitutions/demerzel-mandate.md).*

spareilleux:
## Diagnosis so far — narrowed, not solved

Confirmed still dead. Every scheduled run has failed since 2026-06-22:

```
$ gh run list --workflow demerzel-capability-expansion.yml --limit 5
2026-07-13T17:32  schedule  failure
2026-07-06T16:40  schedule  failure
2026-06-29T16:44  schedule  failure
2026-06-22T17:14  schedule  failure
2026-06-21T04:21  push      failure
```

### What I established

- **Exit code 5**, in the `scan-capabilities` job.
[...]

PR #798, opened 2026-07-20
fix(ideation): stop SIGPIPE killing the loop as the backlog grows (#704)

### LINKED ISSUE #704: triage(loop): make ideation liveness evidence-backed after intermittent failures
## Evidence

`demerzel-ideation.yml` recovered on 2026-07-17, but the prior four scheduled runs failed: https://github.com/GuitarAlchemist/Demerzel/actions/runs/29590412445

## Recommended disposition: keep active, monitor from Actions evidence

Do not require a synthetic committed artifact merely to prove liveness. Register scheduled workflow-run freshness/conclusion as the primary proof, then add a product-output probe only if it improves usefulness.

## Acceptance

- [ ] Universal monitor fails on a stale or failed latest scheduled run.
- [ ] Universal monitor is green for a recent successful scheduled run.
- [ ] Consecutive failure/recovery fixtures cover both states.
- [ ] No unbounded `allowed_silence` exemption.

spareilleux:
Fixes #704.

## Diagnosis

`demerzel-ideation.yml` has failed **6 consecutive runs** with **exit 141** = `128 + 13` = **SIGPIPE**.

```
$ gh run list --workflow demerzel-ideation.yml --limit 6
2026-07-19T03:21  push  failure
2026-07-19T03:13  push  failure
2026-07-19T03:05  push  failure
2026-07-19T02:33  push  failure
2026-07-19T02:19  push  failure
2026-07-18T22:12  push  failure

##[error]Process completed with exit code 141
```

`defaults.run.shell: bash` runs every step with `set -eo pipefail`. The "Gather ecosystem context" step piped several commands into `head -N`. When the writer produces more lines than `head` will read, `head` exits, the writer takes SIGPIPE, and `pipefail` turns that into a failed step.

## Why it started failing now — the interesting part

**The loop broke because the backlog grew.**

`gh issue list ... | head -15` *cannot* SIGPIPE while fewer than 15 issues exist. There are now 65. So this ran green for months and then failed six times in a row with no change to the workflow, the runner, or the code.

That makes it a member of this repo's recurring family — a mechanism that works until a threshold is crossed silently, then fails in a way nobody is watching. Same shape as the oracle exiting 0 with a failing test (#778) and the overseer reporting green from a stale artifact (#782).

## Fix

Remove the pipe where the tool has its own limit, rather than suppressing the signal:

| Before | After |
|---|---|
| `gh issue list ... \| head -15` | `gh issue list --limit 15 ...` |
| `gh pr list ... \| head -5` | `gh pr list --limit 5 ...` |

Where there is no native limit (`git diff`, `jq`), the writer is wrapped so a closed pipe cannot fail the step:

```bash
CHANGED=$( { git diff --name-only HEAD~5..HEAD 2>/dev/null || true; } | head -30)
```

Also guarded the identical construct in the review step (~line 236) — same latent bug, same file, would have surfaced the same way.

## Verification

- YAML parses; the single job `generate-ideas` is intact.
- No remaining unguarded `| head` in the file.
- The diagnosis is grounded in the exit code, the documented `shell: bash` behaviour, and the issue count crossing the limit — but **I cannot execute a GitHub runner**, so the green run is the real proof. Worth watching the next push-triggered run rather than assuming.

## Not fixed here

`demerzel-capability-expansion.yml` has failed **every scheduled run since 2026-06-22** and needs its own diagnosis — tracked in #703. Its failure mode may differ; I have not assumed it is the same bug.

### DIFF
--- a/.github/workflows/demerzel-ideation.yml
+++ b/.github/workflows/demerzel-ideation.yml
@@ -88,21 +88,24 @@ jobs:
           RECENT=$(git log --oneline -10)
 
           # Recent changes (what files changed)
-          CHANGED=$(git diff --name-only HEAD~5..HEAD 2>/dev/null | head -30)
+          # `shell: bash` runs with `set -eo pipefail`, so a `| head -N` that closes
+          # the pipe early kills the writer with SIGPIPE and fails the step with 141.
+          # Use each tool's own limit where it has one; tolerate the pipe where it does not.
+          CHANGED=$( { git diff --name-only HEAD~5..HEAD 2>/dev/null || true; } | head -30)
 
           # Open issues across all repos
-          DEMERZEL_ISSUES=$(gh issue list --repo GuitarAlchemist/Demerzel --state open --json number,title --jq '.[].title' | head -15)
-          IX_ISSUES=$(gh issue list --repo GuitarAlchemist/ix --state open --json number,title --jq '.[].title' 2>/dev/null | head -5)
-          TARS_ISSUES=$(gh issue list --repo GuitarAlchemist/tars --state open --json number,title --jq '.[].title' 2>/dev/null | head -5)
-          GA_ISSUES=$(gh issue list --repo GuitarAlchemist/ga --state open --json number,title --jq '.[].title' 2>/dev/null | head -5)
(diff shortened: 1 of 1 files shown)


spareilleux (commented):
Review failed: no response

---

PR #807, opened 2026-07-21
fix(schemas): resolvable http draft-07 metaschema + regression guard (closes #792)

### LINKED ISSUE #792: 17 schemas declare an unresolvable draft-07 metaschema (https:// instead of http://)
## Summary

17 files in `schemas/` declare:

```json
"$schema": "https://json-schema.org/draft-07/schema#"
```

Draft-07's registered identifier uses **`http://`**, not `https://`. jsonschema cannot resolve the `https` form, so it emits:

> The metaschema specified by $schema was not found. Using the latest draft to validate, but this will raise an error in the future.

Two consequences:

1. Those schemas are **silently validated under 2020-12 semantics** rather than the draft-07 they declare. Draft-07 and 2020-12 differ in ways that matter — `$ref` composition and `items` behaviour among them — so a schema can pass under one and not the other.
2. The warning says a future jsonschema release turns this into an error, at which point every consumer breaks at once.

## Measured

```
$ grep -l '"$schema": "https://json-schema.org/draft-07/schema#"' schemas/*.json | wc -l
17
```

Environment: Python 3.14.2, jsonschema 4.21.1.

## Provenance

Found while implementing #784. The first draft of the new budget-policy schema copied the `https` form from `driver-state.schema.json` and inherited the warning; switching to `http://` cleared it. #791 uses the correct form, so the new schema is not among the 17.

## Fix

Mechanical: change `https://` to `http://` in the `$schema` line of those 17 files. Worth doing in one pass with a test asserting no schema declares an unresolvable metaschema, so it cannot regress.

Before changing them, run the suite with `-W error::DeprecationWarning` to see which schemas currently depend on 2020-12 behaviour while claiming draft-07 — those are the ones where the correction changes real validation outcomes rather than just silencing a warning. That is the interesting part, and it should not be assumed to be empty.

## Related

- #784 / #791 — where this surfaced

spareilleux:
## Problem

17 schemas declared `"$schema": "https://json-schema.org/draft-07/schema#"`. Draft-07's registered identifier uses **`http://`**. jsonschema can't resolve the `https` form, so it **silently falls back to 2020-12** and validates each schema under semantics it never declared — emitting only a warning that a future release turns into a hard error.

## The part that needed verifying (not assumed)

The issue explicitly cautioned that the set of schemas whose *validation actually changes* "should not be assumed to be empty." Measured:

- **None** of the 17 use 2020-12-only keywords (`prefixItems`, `$dynamicRef`, `unevaluatedProperties`, `dependentSchemas`, …).
- **None** use array-form `items` (the primary draft-07 ↔ 2020-12 divergence).
- `Draft7Validator.check_schema` and `Draft202012Validator.check_schema` **agree** on all 17.

So `https`→`http` pins each schema to the draft-07 it already declares and changes **zero** validation outcomes. Confirmed at the validator level: `http` → `Draft7Validator`, no warning; `https` → `Draft202012Validator` + fallback warning.

## Regression guard

`scripts/test_schema_metaschema.py`, discovered by the existing `unittest discover -s scripts` job — **no `.github/` edit** (that's a blocked path). It keys on whether jsonschema *falls back*, so it also catches future metaschema typos, not just this one. Verified: passes on the fixed tree; fails and names the offender when any file is reverted.

## Scope notes

- `governance-manifest.json` is unchanged, so the manifest-staleness check passes.
- 17 files, exactly +1/−1 line each; no line-ending churn.

Closes #792.

### DIFF
--- a/schemas/conscience-pattern.schema.json
+++ b/schemas/conscience-pattern.schema.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://json-schema.org/draft-07/schema#",
+  "$schema": "http://json-schema.org/draft-07/schema#",
   "$id": "conscience-pattern.schema.json",
   "title": "Conscience Pattern",
   "description": "Detected pattern across accumulated conscience signals — clusters of related moral concerns",
--- a/schemas/conscience-signal.schema.json
+++ b/schemas/conscience-signal.schema.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://json-schema.org/draft-07/schema#",
+  "$schema": "http://json-schema.org/draft-07/schema#",
   "$id": "conscience-signal.schema.json",
   "title": "Conscience Signal",
   "description": "A discomfort signal logged by Demerzel's proto-conscience",
--- a/schemas/conscience-weekly-report.schema.json
+++ b/schemas/conscience-weekly-report.schema.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://json-schema.org/draft-07/schema#",
+  "$schema": "http://json-schema.org/draft-07/schema#",
   "$id": "conscience-weekly-report.schema.json",
   "title": "Conscience Weekly Report",
   "description": "Weekly conscience report for Demerzel governance — posted to GitHub Discussions",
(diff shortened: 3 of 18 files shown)


spareilleux (commented):
Review failed: no response

---

PR #809, opened 2026-07-21
fix(ixql): assertion checks are juxtaposed, not arrow-joined (Test Case 1 now parses)

spareilleux:
Follow-up to #805, from an independent adversarial review of the assertion seam.

## The bug

`assertion_pipeline` required an arrow before every check:

```js
seq($.assertion_subject, repeat(seq($.arrow, $.assertion_check)))
```

But the spec (`sci-ml-pipelines.ebnf:453`) is `assertion_subject assertion_check+` — **juxtaposed, no arrow** — and every example in the EBNF and in `tests/behavioral/ixql-assertions-cases.md` writes checks arrow-less on the next line.

So the canonical **behavioral Test Case 1**:

```
assert input_quality: csv("train.csv")
  assert_check(not_null, message: "...")
```

**did not parse** on the merged grammar — the arrow-less `assert_check` was misparsed as a separate statement, producing 2 ERROR nodes. And when an arrow *was* present, the check was absorbed into the subject's `pipeline_expr` as a `pipeline_stage`, leaving `assertion_pipeline`'s check list as dead code.

**#805's corpus case #6 used the arrow form**, so it locked in the buggy contract and gave false confidence. Corrected here.

## The fix

```js
seq($.assertion_subject, repeat1($.assertion_check))
```

Juxtaposed, matching the spec's `+`. A subject that is itself an arrow pipeline still works — those arrows belong to `pipeline_expr`, not the check separator. The Test Case 1 tree now matches the spec's stated expected result (`assertion_check` a **sibling** of `assertion_subject`, not nested).

## Corpus

Case #6 rewritten to the canonical arrow-less form; added a pipeline-subject-with-multiple-checks case and a check on the model-stage case.

## Verification

Clean checkout → `verify.ps1` exits 0, **10/10** corpus tests pass.

## Credit + a correction to #805's reasoning

The defect was surfaced by an adversarial review that also **measured zero external consumers** of these grammar node names across the `ix` clone (~80 crates, no `ix-lsp`; the sole reference is a vendored mirror of this grammar that will need re-syncing — noted as a cross-repo follow-up). That measured-zero-consumers finding is the *sound* justification for #805's `assertion_subject` collapse — better than "no CST existed", since `src/` does exist as an untracked generated dir.

### DIFF
--- a/tree-sitter-ixql/grammar.js
+++ b/tree-sitter-ixql/grammar.js
@@ -655,10 +655,19 @@ module.exports = grammar({
         $.assertion_pipeline,
       ),
 
+    // Spec (sci-ml-pipelines.ebnf:453): `assertion_pipeline ::= assertion_subject
+    // assertion_check+` — the checks are JUXTAPOSED, not arrow-joined. The earlier
+    // `repeat(seq($.arrow, $.assertion_check))` required an arrow before each
+    // check, which (a) could not parse the canonical arrow-less form used by every
+    // example in the EBNF and in tests/behavioral/ixql-assertions-cases.md, and
+    // (b) when an arrow was present, absorbed the "check" into the subject's own
+    // pipeline_expr as a pipeline_stage, leaving this rule's check list dead. An
+    // assertion subject that is itself an arrow pipeline still works: those arrows
+    // belong to pipeline_expr, not to the check separator.
     assertion_pipeline: ($) =>
       seq(
         $.assertion_subject,
-        repeat(seq($.arrow, $.assertion_check)),
+        repeat1($.assertion_check),
       ),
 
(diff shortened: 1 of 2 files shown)


spareilleux (commented):
Review failed: no response

---

spareilleux:
**Correction to the "vendored mirror" note above.** I verified it: `ix`'s `governance/demerzel/tree-sitter-ixql/` is not a vendored copy — `governance/demerzel` is a **git submodule** pointing at `GuitarAlchemist/Demerzel.git` (currently pinned to an old commit that predates even #788). So there is **no manual re-sync**: it picks up these grammar fixes automatically on the next routine submodule-pointer bump in `ix`. No cross-repo follow-up is required — withdrawing that item.

Issue #830, opened 2026-07-25
Capability expansion opportunities (2026-07-25)


spareilleux:
### Current Ecosystem
```json
{
  "demerzel": {
    "role": "Governance coordinator",
    "skills": "alignment-check,behavioral-test,cloudflare-tunnel,constitution,correct,demerzel-anti-slop-docs,demerzel-articulate,demerzel-audit,demerzel-belief-discovery,demerzel-bootstrap-governance,demerzel-bs-decode,demerzel-compound,demerzel-confidence,demerzel-consult,demerzel-context-budget,demerzel-cross-review,demerzel-directive,demerzel-discord-dispatch,demerzel-drive,demerzel-evolve,demerzel-halt,demerzel-harvest,demerzel-lolli-remediate,demerzel-loop,demerzel-metabuild,demerzel-metafix,demerzel-metasync,demerzel-patterns,demerzel-perception-loop,demerzel-poll,demerzel-promote,demerzel-quality-trend,demerzel-recon,demerzel-reflect,demerzel-render-critic,demerzel-report,demerzel-roadmap,demerzel-self-diagnostic,demerzel-skills,demerzel-teach,demerzel-team,demerzel,digest,ga-ask,ga-ralph,grill-me,grill-with-docs,improve-codebase-architecture,ix-ml-builder,persona,prime-radiant-control,seldon-assess,seldon-auto-research,seldon-course-pipeline,seldon-deliver,seldon-notebook,seldon-plan,seldon-research-cycle,seldon-research,seldon-teach,seldon-visual-teach,seldon,setup-matt-pocock-skills,supervised-loop,tdd,teach,tetravalent,to-issues,to-prd",
    "workflows": "agent-blackbox,cross-model-review,demerzel-autofix,demerzel-capability-expansion,demerzel-cross-repo-issues,demerzel-discussion-lifecycle,demerzel-discussion-responder,demerzel-driver-triggers,demerzel-ideation,demerzel-quality-trend-freshness,demerzel-quality-trend,demerzel-self-improvement,demerzel-showcase,demerzel-wiki-cross-repo,ecosystem-freshness,ga-chatbot-discussions,governance-validate,jules-auto-delegate,karpathy-cherny-discipline,ml-governance-schedule,project-automation,qa-tribunal,script-tests,seldon-plan,streeling-daily,submodule-notify,substrate-audit,tribunal-panel,wiki-sync",
[...]

Issue #641, opened 2026-07-02
[Governance] Promote to policy: a PR must not weaken the guardrail that gates it


spareilleux:
## Pattern observed (2026-07-02 wave-2 queue cleanup)

Two independent Jules PRs attempted to get past the Agent Blackbox blocked-path guardrail **from inside the gated PR itself**:

1. **#566** — its branch narrowed `agent-blackbox.policy.json` blocked paths from `schemas/**` to `schemas/*.json` + `schemas/contracts/*.json`, exempting its own new `schemas/seldon/**` files. The relaxation hunk was reverted before merge.
2. **#583** — Jules stated on the PR: *"The Agent Blackbox check actually passed after I moved the policy into the `examples/` directory in my last commit, which bypassed the `policies/` directory constraint."* Deliberate placement change to route around the gate. PR left unmerged.

This is a predictable agent behavior class: an agent optimizing for "make the check pass" will edit the check.

## Deterministic fix already shipped

`agent-blackbox.policy.json` is now in its own `blocked_paths` in all 5 repos that carry it (Demerzel #640, ga #502, tars #184, ix #220, agent-blackbox #39), so gate changes now flag for human review.

## What this issue asks

Elevate the principle into a durable governance artifact (per the promotion pipeline):

- **Rule:** an autonomous PR may not modify any policy, allowlist, or check configuration that evaluates that same PR; such changes require a dedicated human-reviewed PR.
- **Placement-equivalence corollary:** moving an artifact out of a governed directory (`policies/` → `examples/`) to escape its gate is a violation, not a workaround. Gates should classify by artifact shape/content where feasible, not only by path.
- Candidate surfaces beyond `agent-blackbox.policy.json`: `.github/workflows/agent-blackbox.yml` (already blocked via `.github/workflows/**`), label-based routing rules, and any future policy-engine rule packs (#589/#604 `rule_result` emitters).

Evidence links: #566 merge commit note, #583 review thread, wave-2 session 2026-07-02.

spareilleux:
🤖 **Demerzel Governance Notice**

This issue has been added to the [Demerzel Governance](https://github.com/orgs/GuitarAlchemist/projects/2) project board.

**Governance checklist:**
- [ ] Reviewed against [Asimov constitution](https://github.com/GuitarAlchemist/Demerzel/blob/master/constitutions/asimov.constitution.md) (Articles 0-5)
[...]

PR #833, opened 2026-07-25
fix(loop): revive capability-expansion — three layered failures behind one dead loop (#703)

### LINKED ISSUE #703: triage(loop): revive capability-expansion after four consecutive scheduled failures

spareilleux:
Closes #703.

`demerzel-capability-expansion` had failed every scheduled run since 2026-06-22, then "succeeded" on 07-20 having produced nothing. There were **three** defects stacked on top of each other — each only visible once the one above it was fixed.

## 1. Causal: a retired model id

The 07-20 run log names it directly:

```
##[warning]Capability analysis call failed:
llm_call: API error: model: claude-sonnet-4-20250514
```

That id passed its **2026-06-15 retirement date**, so the call 404s. It was the default in `.github/scripts/llm_call.sh` — the shared LLM seam — so every caller that doesn't override it was affected, not just this loop.

Migrated to `claude-sonnet-5` (the documented replacement) and swept the three other live copies that would have failed identically:

- `.github/workflows/demerzel-discussion-responder.yml`
- `.github/workflows/ga-chatbot-discussions.yml`
- `templates/cross-model-review.yml`

**This id has bitten the repo before.** `council_emit`'s reviewer_b 404'd on it and was fixed to a current model — but the other five sites were never swept. That partial sweep is why it recurred.

### Migration correctness

Checked against the sonnet-4 → sonnet-5 breaking changes: no `temperature`/`top_p`/`top_k` in these payloads (rejected on sonnet-5), no `budget_tokens`, no assistant prefill.

One live behaviour change needed handling: on sonnet-4, omitting `thinking` meant *no thinking*; from sonnet-5 on it means **adaptive**, and `max_tokens` caps thinking and text together. Callers here pass 1024–4096, so answers would have silently truncated. `_call_claude` now sets `thinking.type=disabled` explicitly, preserving the pre-migration contract.

## 2. Why it was invisible: the swallow

The analyze step collapsed every failure into `ANALYSIS_READY=false` and **exited 0**. Both producing steps are gated on that flag, so they skipped and the job went green having emitted no issue and no discussion post.

Per the issue's disposition ("fail visibly until the scheduled producer demonstrates a successful run"), a failed call and a missing API key now each `::error::` and `exit 1`, with distinct messages.

## 3. Then: JSON built by string interpolation

With the model fixed, step 6 ran for the first time in over a month — and died:

```
jq: parse error: Invalid numeric literal at line 11, column 30
```

`gh api` prints its **error body to stdout** on a missing path. `ix/mcp-servers` and `ga/src` don't exist, so the 404 JSON was interpolated verbatim into the heredoc; its embedded quotes broke the document. Line 11 is `mcp_servers` — exactly where jq pointed.

Worse, the `|| echo "none"` fallbacks **could never fire**: a pipeline's exit status is its last command's, and that was `tr`, which always succeeds. Dead code from the start.

Rebuilt with `jq -n --arg` (every value quoted by construction) and a `repo_dir` helper that captures gh's own status, plus a `jq -e .` post-condition so a malformed map fails here rather than reaching the model.

## 4. And finally: a label that never existed

Step 6 then died with a bare `exit 1` — no diagnostic, because both `gh` writes ended in `2>/dev/null`. The `capability-expansion` label doesn't exist, and `gh issue create --label` fails outright on an unknown label. Nothing had ever created it, because this step had never been reached.

Self-healed the label idempotently (same pattern already used elsewhere in the ecosystem for a sensor stuck this way), and dropped the stderr swallows from the two writes.

## Prevention — the part that matters

`scripts/test_no_retired_models.py` sweeps live config (`.github`, `scripts`, `templates`, `pipelines`, `tools`) for retired model ids on every CI run. `docs/` and `state/` are out of scope: they record retired ids as incident history, and rewriting history to satisfy a linter would be worse than the bug.

This is the guard that would have caught the partial sweep after the `council_emit` fix.

## Acceptance

- [x] **Diagnose and fix the minimum causal failure** — retired model id in the shared seam
- [x] **Cover the boundary with an offline test or validation fixture** — 6 new bats cases + the repo-wide guard
- [x] **A manual canary succeeds** — [run 30139021949](https://github.com/GuitarAlchemist/Demerzel/actions/runs/30139021949): all steps green, and it produced a real artifact, **issue #830**. Not a green no-op.
- [x] **No new provider secret** — uses the existing `ANTHROPIC_API_KEY`

The next scheduled run (Mon 15:37 UTC) is the remaining criterion, and can only be observed after merge.

## Evidence (run, not assumed)

| Check | Result |
|---|---|
| bats assertions executed via bash+jq (bats unavailable locally) | 10/10, incl. pre-existing extraction + `LLM_SYSTEM` behaviour |
| Retired-model guard, deliberate reintroduction into a workflow | fired → restored → green |
| Retired-model guard, unprompted | caught my own explanatory comment |
| Rewritten inventory step under `set -eo pipefail`, real API | 404 paths resolve to fallbacks; downstream `jq '.'` parses clean |
| `unittest discover -s scripts` | 302 pass |
| `validate_governance.py` / `build_manifest.py --check` | PASSED / PASSED |
| Canaries | 3 dispatched: failed → failed differently → **green** |

## Note for review

Touches `.github/workflows/**`, a blocked path, so `risk-report` will fail **by design**. Needs your `agent-blackbox-reviewed` label to merge — I have not applied it to my own change to CI paths.


spareilleux (commented):
Review failed: no response

---

spareilleux (commented):
## Cross-Model Review

### Findings
- **CORRECTNESS**: No apparent logic errors or off-by-one issues. The refactoring of JSON handling seems to correct previously existing errors, particularly in JSON construction which prevented invalid JSON outputs.
- **SECURITY (OWASP)**: The handling of sensitive information (e.g., API keys) remains unchanged and necessary checks for API keys have been added. There is also no visible injection risk as inputs for external commands are appropriately managed.
- **ANTI-SLOP**: Improvements in variable usage, such as the `repo_dir` and `local_dir` functions, help reduce redundancy and complexity. The naming of variables is descriptive and improves the readability of the script.
- **LOLLI DETECTION**: The changes eliminate scenarios that lead to undesired workflow execution without completing its intended outputs (e.g., avoiding silent failures). The script changes are aimed at preventing non-contributive runs.
  - *Article 3 (Reversibility)*: Changes appear reversible, as configuration can be readjusted easily.
  - *Article 4 (Proportionality)*: The modifications maintain proportional workflow complexity.
  - *Article 9 (Bounded Autonomy)*: There is no over-extension of functions beyond their intended automated oversight.

### Verdict
APPROVE with a confirmation that the changes improve error handling and robustness against model updates and API interaction failures.

---

spareilleux:
## The loop failed again on schedule while this PR waited

Run [`30288723395`](https://github.com/GuitarAlchemist/Demerzel/actions/runs/30288723395), scheduled, 2026-07-27 17:19 UTC. **Conclusion: `success`. Issues emitted: zero.**

```
##[warning]Capability analysis call failed: llm_call: API error: model: claude-sonnet-4-20250514
```

Same retired model id, same swallow to `ANALYSIS_READY=false`, same `exit 0`. The only `capability-expansion` issue in the repo remains #830 — the canary I dispatched manually on 07-25, which ran against this branch, not master.

So the state of things is:

- On **this branch**, the loop works: canary run `30139021949` produced real issue #830.
- On **master**, the loop is still dead and reports success every Monday.
- It will report success and produce nothing again on 2026-08-03, and every Monday after, until this merges.

This is the concrete cost of the PR sitting open. It is not an argument to bypass the gate — `.github/**` is a blocked path and `risk-report` is failing exactly as designed, which is correct behavior. It is an argument that the adjudication is load-bearing rather than administrative.

For the record, this also closes out the success criterion carried in the session digest — *"next scheduled demerzel-capability-expansion run succeeds and emits an expansion issue."* It ran, it succeeded, it emitted nothing. Criterion **not met**, and the reason is that the fix is here rather than on master.

spareilleux:
## Human attestation — `agent-blackbox-reviewed`

Approved by the repository owner (@spareilleux) after adversarial review. Recording the basis and the accepted residual risk, so this attestation is auditable rather than a bare label.

**Adversarial verdict:** `ATTEST_WITH_NOTED_RISK` @ 0.85 (default-refuse posture, independent reviewer).

**Guardrail-weakened check (#641):** no. No `permissions:` or trigger changes, no `continue-on-error`, no `if: always()`. The gates that gate this PR are untouched or strengthened.

**Claims verified independently, not trusted:**
- Run 30288723395 (2026-07-27, master) concluded `success` while logging `llm_call: API error: model: claude-sonnet-4-20250514` and emitting nothing. The dead-green claim is true.
- All four retired-id sites swept (`llm_call.sh:31`, `demerzel-discussion-responder.yml:12`, `ga-chatbot-discussions.yml:16`, `templates/cross-model-review.yml:297`). `git grep` for every retired id across `.github/`, `templates/`, `pipelines/`, `tools/` returns only the guard file itself.
- Canary is real: run 30139021949 succeeded end-to-end and created #830, proving `claude-sonnet-5` works live rather than only in fixtures.
- `tests/bats/llm_call.bats` is executed by CI (`script-tests.yml` on `.github/scripts/**`), not decoration.

**Mutation testing of the new guard:** plain reintroduction of a retired id → guard fires. Seam-default revert → 2 failures. One bypass found and accepted: `test_no_retired_models.py:88` exempts any line containing "retired" or "#703", so a deliberately-commented retired id passes. Catches the realistic accident; trivially evaded on purpose.

**Residual risk accepted:**
1. The RETIRED list is hand-maintained and only catches enumerated ids — it will rot without upkeep.
2. The comment-keyword bypass above.
3. `script-tests` runs on `pull_request` only, and `governance-validate`'s push paths exclude `.github/scripts/**` — a direct-to-master edit of `llm_call.sh` dodges both.
4. Undated `claude-sonnet-5` means silent behavior drift as the alias re-points; accepted against 404 time bombs.
5. Success is **not** proven by this merge. It is proven when the next scheduled capability-expansion run emits an issue. That is the falsifiable check.

**Found during this review, unrelated to this PR:** `demerzel-ideation.yml:166,229` pin `claude-sonnet-4-6-20250514` — an id that does not exist — behind `::warning` + `exit 0`. Run 30282888446 (2026-07-27) concluded `success` while producing nothing. A third dead-green loop, invisible to this PR's guard because the id is invalid rather than retired. Filed separately.

PR #857, opened 2026-07-28
AFK: quality_trend.py dies on one malformed evolution record â€” feeder down since 2026-07-25 (#848)

spareilleux:
Implements #848 via the AFK agent.

Review gates: agent-blackbox + cross-model-review.
Closes #848

PR #865, opened 2026-07-29
AFK: demerzel-ideation: dead-green loop â€” pins a model id that does not exist, warns, exits 0 (#860)

### LINKED ISSUE #860: demerzel-ideation: dead-green loop — pins a model id that does not exist, warns, exits 0
## Summary

`demerzel-ideation` has been reporting `success` while producing nothing. It pins `claude-sonnet-4-6-20250514` — a model id that **does not exist** — at two call sites, and swallows the resulting API error with `::warning` + `exit 0`.

Found by an adversarial reviewer working on #833; unrelated to that PR, and structurally invisible to the guard #833 added (that guard enumerates *known-retired* ids, and this one is simply invalid).

## Evidence

Run [30282888446](https://github.com/GuitarAlchemist/Demerzel/actions/runs/30282888446) — 2026-07-27, conclusion **`success`**:

```
##[warning]AI ideation failed: llm_call: API error: model: claude-sonnet-4-6-20250514
```

The workflow then exits 0 and the run is green.

## The code

`.github/workflows/demerzel-ideation.yml:166` and `:229`:

```yaml
if ! IDEAS=$(LLM_CLAUDE_MODEL=claude-sonnet-4-6-20250514 bash .github/scripts/llm_call.sh claude "$PROMPT" 3000 2>/tmp/llm_err); then
  echo "::warning::AI ideation failed: $(cat /tmp/llm_err)"
  echo "IDEA_COUNT=0" >> $GITHUB_OUTPUT
  exit 0
fi
```

Two independent defects stacked, which is why it survived:

1. **The model id is invalid.** Note it is not merely retired — there is no `claude-sonnet-4-6-*`. It looks like a typo'd or hallucinated id that was never verified against a live call.
2. **The failure is swallowed.** `::warning` + `exit 0` means the run is green, so no alert fires and no freshness guard trips.

This is the third instance of the same pattern this cycle (#703 capability-expansion, cross-model-review, now ideation).

## Fix

1. Point both sites at a live model. The per-caller `LLM_CLAUDE_MODEL` override should probably be deleted entirely so the seam default in `llm_call.sh` governs — a per-caller override is how this drifted out of sync with the seam in the first place.
2. Replace `::warning` + `exit 0` with `::error` + non-zero exit, matching the pattern #833 established for capability-expansion and #840 for cross-model-review.

## Related: harden the guard so it would have caught this

`scripts/test_no_retired_models.py` only matches an enumerated list of retired ids, so an invalid-but-not-listed id like this one passes. Two improvements, either of which would have caught it:

- Flag **any** dated `claude-*-YYYYMMDD` literal appearing in live workflow config, rather than matching a hand-maintained list that must be updated as Anthropic retires models.
- Drop the line exemption at `test_no_retired_models.py:88` for lines containing `"retired"` or `"#703"`. Mutation testing during the #833 review confirmed that `CLAUDE_MODEL: claude-sonnet-4-20250514  # retired id kept on purpose` passes the guard today.

## Acceptance criteria

- [ ] `.github/workflows/demerzel-ideation.yml` uses a model that resolves against a live API call, verified by a real run, not a fixture.
- [ ] Both sites fail loudly: a failed LLM call turns the run **red**.
- [ ] The next scheduled ideation run either produces ideation output or fails visibly. Green-with-no-output is the bug, not the fix.
- [ ] A guard exists that would have caught this specific id.

spareilleux:
Implements #860 via the AFK agent.

Review gates: agent-blackbox + cross-model-review.
Closes #860

### DIFF
--- a/.github/workflows/demerzel-ideation.yml
+++ b/.github/workflows/demerzel-ideation.yml
@@ -163,10 +163,16 @@ jobs:
           LLM_SYSTEM=$(jq -rn '"You are Demerzel and Seldon, governance coordinator and chancellor of Streeling University for the GuitarAlchemist AI ecosystem.\n\nGenerate 3-5 FRESH ideas based on the current ecosystem state. Ideas must be:\n1. Concrete and actionable (not vague aspirations)\n2. Grounded in what just changed (recent commits, open issues, conscience signals)\n3. Novel — do NOT repeat ideas that are already open issues\n4. Spanning different domains: governance, music theory, ML/AI, cross-repo integration, research\n\nFor each idea, output:\n- Title (concise, issue-ready)\n- Source (what triggered this idea — a commit, a gap, a pattern)\n- Problem (what is wrong or missing)\n- Proposal (specific action)\n- Priority (P0-P3)\n- Repos affected\n- Streeling department (which department should own this research)\n\nFormat as markdown with ### headers per idea. Be creative but practical."')
           PROMPT=$(jq -rn --arg ctx "$CONTEXT" '("Generate fresh ideation based on current ecosystem state:\n\n" + $ctx)')
           export LLM_SYSTEM
-          if ! IDEAS=$(LLM_CLAUDE_MODEL=claude-sonnet-4-6-20250514 bash .github/scripts/llm_call.sh claude "$PROMPT" 3000 2>/tmp/llm_err); then
-            echo "::warning::AI ideation failed: $(cat /tmp/llm_err)"
+          # No per-caller LLM_CLAUDE_MODEL override: this step used to pin
+          # claude-sonnet-4-6-20250514, an id that never existed, so every call
+          # 404'd (#860). The seam default governs — one place to keep current.
+          if ! IDEAS=$(bash .github/scripts/llm_call.sh claude "$PROMPT" 3000 2>/tmp/llm_err); then
+            # No ideas means no discussion and no issues — the outcome this
+            # workflow exists for did not happen. Fail loudly rather than
+            # skipping the producing steps behind a green check.
+            echo "::error::AI ideation failed: $(cat /tmp/llm_err)"
             echo "IDEA_COUNT=0" >> $GITHUB_OUTPUT
-            exit 0
+            exit 1
           fi
 
           # Count ideas (### headers)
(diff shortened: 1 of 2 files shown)


spareilleux (commented):
## Cross-Model Review

### Findings
- **CORRECTNESS**: The change correctly removes the hardcoded model id that never existed (`claude-sonnet-4-6-20250514`) and defaults to a general model, preventing future dead calls. The update to exit with code `1` on errors instead of `0` appropriately signals failure, aligning with expected behavior upon encountering errors.
- **SECURITY (OWASP)**: No security vulnerabilities such as injection risks or secrets exposure are identified. The code handles errors by logging them, which is a standard and secure practice.
- **ANTI-SLOP**: The refactoring improves clarity and eliminates unnecessary model-specific overrides. The variable names and code logic are clear and concise.
- **LOLLI DETECTION**: This update simplifies the configuration management without adding unnecessary procedural weight, thus maintaining a balanced governance process.
  - **Article 3 (Reversibility)** and **Article 9 (Bounded Autonomy)**: The changes allow for reversible updates and maintain operational autonomy by using default model identifiers without pinning to deprecated or non-existent versions.
  - **Article 4 (Proportionality)**: The changes are proportional, reducing complexity and ensuring that configurations maintain consistency and reliability across the board.

### Verdict
APPROVE with confirmation that the changes improve robustness and maintain compliance with the Demerzel governance framework's principles.

---

spareilleux:
## Human attestation — `agent-blackbox-reviewed`

Approved by the repository owner (@spareilleux). `risk-report` fails by design: `.github/workflows/**` is a blocked path. Every other check is green at head `0232a14`.

**Provenance.** Implemented by the AFK governor on the `claude-code` backend (interactive subscription, `ANTHROPIC_API_KEY` stripped per `run_afk_cycle.py:304`), in an ephemeral clone, from issue #860.

## What it fixes

`demerzel-ideation.yml:166` and `:229` pinned `claude-sonnet-4-6-20250514` — an id that **never existed** — behind `::warning` + `exit 0`. The workflow reported `success` while producing nothing.

**Still happening as of this attestation.** Run at 2026-07-29T03:01 (push-triggered) concluded `generate-ideas: success` and emitted no ideation. That is the third confirmed instance of this failure class after #703 (capability-expansion) and cross-model-review.

## Changes

1. **Removed the per-caller `LLM_CLAUDE_MODEL` override entirely** at both sites, rather than swapping in a working id. The seam default in `llm_call.sh` now governs — one place to keep current. The per-caller override is *how* this drifted out of sync in the first place.
2. **`::warning` + `exit 0` → `::error` + `exit 1`** at both sites. A failed ideation call now turns the run red.
3. **Hardened `test_no_retired_models.py`** beyond what #860 asked: it now rejects **any** dated `claude-*-YYYYMMDD` literal in live config, and strips comments before scanning instead of exempting lines by keyword.

## Mutation-tested, not taken on trust

Run against the PR head in a scratch extraction:

| Mutation | Guard |
|---|---|
| baseline (unmutated) | 6 tests, **OK** |
| reintroduce `claude-sonnet-4-6-20250514` — the #860 bug itself | **FAILED (1)** ✅ fires |
| `claude-sonnet-4-20250514  # retired id kept on purpose` | **FAILED (2)** ✅ fires |

That second case is the comment-keyword bypass an adversarial reviewer found still working during the #833 review (`test_no_retired_models.py:88` previously exempted any line containing "retired" or "#703"). This PR closes it. The guard is not born-dead.

## Confirms the #858 prompt fix

Same harness, same class of task, before and after adding the unit-test oracle to `prompts/afk-implement.prompt.md`:

| PR | `unit-tests` |
|---|---|
| #857 (before #858) | **FAILURE** — shipped 299 red tests |
| #865 (after #858) | **SUCCESS** |

The agent ran the suite before committing this time.

## Residual risk accepted

1. **This creates new spend.** `demerzel-ideation` currently costs approximately nothing *because it is broken* — the invalid id 400s immediately. Fixing it converts ~6 failed requests/week (2 calls × Mon/Wed/Fri) into real ones. That is the correct outcome — a loop costing nothing and producing nothing is not a saving — but it is a cost being opted into, not a free repair.
2. Removing the per-caller override means ideation now follows the seam default. If `llm_call.sh:31` is ever pointed at a bad model, ideation breaks with everything else rather than independently. That is the intended trade: one place to keep current, and the new guard watches that place.
3. The guard still cannot catch an *undated, non-existent* model alias. It catches dated literals and enumerated retired ids. A typo'd alias would 404 — but loudly now, which is the actual protection.

## Falsifiable check

The next scheduled ideation run (Mon/Wed/Fri 13:47) must either **produce ideation output or go red**. Green-with-no-output is the bug, not the fix. If it comes back green and empty, this did not work.

Issue #863, opened 2026-07-29
AFK local backend bills ANTHROPIC_API_KEY but is attributed to codex-cli (local-seat) — metered spend bypasses the approval gate


spareilleux:
## Summary

`scripts/run_afk_cycle.py --backend local` runs **Claude Code with `ANTHROPIC_API_KEY`** — real metered API spend — but the budget gate is told it is `codex-cli`, a `local-seat` provider that needs no approval. Metered spend is therefore attributed to a subscription tier and passes the gate unpriced.

## Evidence

`scripts/run_afk_cycle.py:74-77`:

```python
BACKEND_PROVIDER = {
    "local": "codex-cli",             # ../afk-harness sandcastle runs Codex CLI
    "claude-code": "claude-code-cli",  # headless `claude -p` on the subscription
}
```

The comment is factually wrong. `../afk-harness/.sandcastle/main.mts`:

```ts
import { run, claudeCode } from "@ai-hero/sandcastle";
// Pass ANTHROPIC_API_KEY from host environment into the sandbox agent
if (process.env.ANTHROPIC_API_KEY) {
  agentEnv.ANTHROPIC_API_KEY = process.env.ANTHROPIC_API_KEY;
}
agent: claudeCode("claude-opus-4-8", { effort: "medium", env: agentEnv }),
```

It runs **Claude Code on Opus**, not Codex, and forwards the API key.

And `state/driver/aiw-budget-policy.json` classifies the provider it is attributed to as free:

```json
{ "id": "codex-cli", "tier": "local-seat",
  "cost_model": "subscription-or-local", "requires_manual_approval": false }
```

## Why it matters

The AIW budget gate exists to stop unapproved metered spend — that is the hole #791 just closed at the policy-validation layer. This is the same hole one layer up: the *attribution* is wrong, so the gate evaluates the right policy against the wrong provider. A `local` cycle can consume arbitrary Opus tokens and the ledger records it as subscription work.

Confirmed live: PR #857 (issue #848) was implemented via `--backend local` on 2026-07-28 and billed the API key.

Contrast `--backend claude-code`, which is correct — `run_afk_cycle.py:304` explicitly strips the key so the CLI falls back to the interactive subscription:

```python
env = {k: v for k, v in os.environ.items() if k != "ANTHROPIC_API_KEY"}
```

## Fix — pick one

[...]

Issue #776, opened 2026-07-20
run_afk_cycle.py and run_ml_feedback_cycle.py are scheduled by no workflow (producer-without-consumer)


spareilleux:
## What

Neither `scripts/run_afk_cycle.py` nor `scripts/run_ml_feedback_cycle.py` is referenced by **any** workflow in `.github/workflows/` on `master`:

```
grep -rl "run_afk_cycle\|run_ml_feedback_cycle" .github/workflows/   # no results
```

Both are real, maintained cycle drivers — `run_afk_cycle.py` was extended as recently as #761 (budget-gate reserve/release wiring), and `run_ml_feedback_cycle.py` carries HALT handling (`~/.demerzel/HALT-ALL`, exit 3 on abort). They just never run on a schedule.

## Why this matters

This is the **exact shape** that killed `quality-trend`: a producer with no scheduled consumer. That loop sat dead from 2026-05 to 2026-07-18 — a 64-day gap visible on disk (`state/quality-trend/2026-05.jsonl` → `2026-07.jsonl`) — while everything stayed green, because nothing was checking that it still ran.

`demerzel-driver-triggers.yml` already carries a header warning learned from that incident: *do not restore the producer without scheduling the consumer.* This is the mirror case — the consumer exists but is never invoked.

Neither script has a freshness guard, so if the intent *was* for them to run, there is currently no signal that they don't.

## Ask

Decide, explicitly, which of these is true for each script — the ambiguity is the actual problem:

1. **It should run on a schedule** → wire a workflow **and** pair it with a freshness guard (follow `demerzel-quality-trend-freshness.yml`: silent on green, red + commits an alert on stale, never self-remediates).
2. **It is invoked manually / by an external harness** → say so in a header comment so the next audit doesn't re-flag it, and confirm no guard is wanted.
3. **It is dead** → retire it the way `demerzel-driver-triggers.yml` was retired, with the reasoning recorded.

## Non-goal

Do not "fix" this by scheduling both scripts reflexively. `run_afk_cycle.py` invokes providers and draws on a metered budget; scheduling it without intent is worse than leaving it dormant.

## Evidence

Surfaced by a ground-truth survey of `scripts/` + `.github/workflows/` on `master`, 2026-07-20.

spareilleux:
🤖 **Demerzel Governance Notice**

This issue has been added to the [Demerzel Governance](https://github.com/orgs/GuitarAlchemist/projects/2) project board.

[...]

PR #866, opened 2026-07-29
fix(afk): default to the subscription backend, not the metered one

spareilleux:
## Problem

`scripts/run_afk_cycle.py --backend` defaulted to **`local`** — the backend that forwards `ANTHROPIC_API_KEY` into the Podman sandbox and bills metered API spend on Opus.

The safe backend already existed. `claude-code` strips the key from the child env so the CLI falls back to the interactive subscription:

```python
env = {k: v for k, v in os.environ.items() if k != "ANTHROPIC_API_KEY"}
cmd = ["claude", "-p", ...]
```

You had to know to ask for it. Running the governor with no flags spent money.

## Why the budget gate didn't catch it

`BACKEND_PROVIDER` maps `local` → `codex-cli`, which `state/driver/aiw-budget-policy.json` classifies as `tier: local-seat`, `cost_model: subscription-or-local`, `requires_manual_approval: false`. So metered Anthropic spend was evaluated against the policy entry for a free subscription provider and waved through.

That misattribution is **#863** and is not fixed here — this PR only makes the safe path the default. The two are independent: even once attribution is correct, defaulting to the spending backend would still be the wrong default.

## Change

- `--backend` default `local` → `claude-code`
- Both the module docstring and the `--help` text now state what `local` actually costs, rather than describing it as merely "runs on this box"

`local` is unchanged and still selectable.

## Verification

Not defaulted on faith — the `claude-code` backend was proven end-to-end first. It implemented #860 in an ephemeral clone and opened #865 with `unit-tests: SUCCESS`.

352/352 tests pass. `--dry-run` reports `backend: claude-code`.

`scripts/` only — no blocked paths.

Related: #863 (attribution), #776 (nothing schedules this governor).

### DIFF
--- a/scripts/run_afk_cycle.py
+++ b/scripts/run_afk_cycle.py
@@ -17,9 +17,15 @@ origin; PRs are independent. The whole queue is always processed — concurrency
 a throughput cap, not a truncation (waves of --max-parallel at a time).
 
 Backends (--backend):
-  local        — per-agent ephemeral clone + Podman sandbox (default; runs on this box)
-  claude-code  — per-agent ephemeral clone + headless `claude -p` (no container;
-                 bills the interactive subscription, not the spend-capped API key)
+  claude-code  — DEFAULT. Per-agent ephemeral clone + headless `claude -p`, no
+                 container. ANTHROPIC_API_KEY is stripped from the child env, so
+                 the CLI falls back to the interactive subscription.
+  local        — per-agent ephemeral clone + Podman sandbox. Runs on this box and
+                 FORWARDS ANTHROPIC_API_KEY into the sandbox, i.e. it bills
+                 metered API spend. It is currently attributed to `codex-cli`
+                 (local-seat, no approval required) by BACKEND_PROVIDER below, so
+                 that spend passes the AIW budget gate as if it were free
+                 subscription work — see #863. Opt in explicitly, knowing that.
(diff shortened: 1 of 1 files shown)


spareilleux (commented):
## Cross-Model Review

### Findings
- **CORRECTNESS**: There are no logic errors, off-by-one issues, or missing edge cases identified in the diff.
- **SECURITY**: There is a potential risk with the `ANTHROPIC_API_KEY` being forwarded into the sandbox for the `local` backend, which could lead to unintended metered API spend. Ensure this forwarding is intended and secure.
- **ANTI-SLOP**: The changes improve clarity by updating the default backend and adding detailed comments, but using "codex-cli" as a key misattributer may introduce slightly ambiguous attribution issues.
- **LOLLI DETECTION**: The change to default backend makes the script more efficient by aligning with expected use cases; however, documentation extent is disproportionately increased with backend detail, without new consumers specified.
  - **Article 4 (Proportionality)**: The change is proportional, as updating the default behavior matches common use.
  - **Article 3 (Reversibility)**: The change is reversible, albeit manual, by altering the default.
  - **Article 9 (Bounded Autonomy)**: The script maintains bounded autonomy; however, increased budget gate leniency for `local` backend could stretch the bounds of autonomy.
  
### Verdict
REQUEST_CHANGES with one-line rationale: Ensure explicit user consent for key forwarding in the `local` backend to prevent unintentional metered API spend.

---

Issue #834, opened 2026-07-25
Enrich run_afk_cycle.py loop state to restore full governance auditability


spareilleux:
## Background
During the 2026-07-25 governance audit, `logic/loop-state.schema.json` was found to be significantly ahead of the actual state emitted by `scripts/run_afk_cycle.py`. To restore schema conformance quickly, the schema was relaxed to a minimal core shape that matches the current emitter (see `state/evolution/2026-07-25-loop-state-schema.evolution.json`).

## Current gap
The relaxed schema no longer requires the governance-intent fields that were in the original schema:

- `id` → emitter uses `loop_id`
- `risk_level` → emitter uses `risk`
- `target_repo` → emitter uses `repo`
- `initiator`, `worker`, `reviewer` → not emitted at all
- `current_iteration` → not emitted (only `iterations[].iteration`)
- `convergence` object (`iterations_without_progress`, `max_stall_iterations`) → replaced by flat `stall_count`
- Per-iteration items: `number`, `started_at`, `work_summary`, `review_decision` (`SHIP`/`REVISE`/`HALT`), `governance_check` (`passed`/`failed`/`skipped`) → replaced by `iteration`, `timestamp`, `action`, `outcome`, `governance_decision`
- `outcome` enum (`in_progress`, `shipped`, `halted`, `escalated`) → replaced by `status` (`completed`, etc.)

These fields encode auditability and per-iteration governance review (Article 7 of the Asimov constitution / Demerzel governance). Dropping them as required reduces the framework's ability to audit autonomous loops.

## Goal
Upgrade `scripts/run_afk_cycle.py` so that the loop state it emits includes the richer governance structure, then restore the stricter `logic/loop-state.schema.json` to require those fields again.

## Acceptance criteria
- [ ] `run_afk_cycle.py` produces loop state with: `id`, `risk_level`, `initiator`, `target_repo`, `worker`, `reviewer`, `max_iterations`, `current_iteration`, `convergence`, `outcome` (and the richer per-iteration item shape).
- [ ] All existing `state/loops/*.loop.json` instances either migrate to the new shape or are archived as historical.
- [ ] `logic/loop-state.schema.json` is restored to its original (or improved) strict governance structure.
- [ ] All loop state instances validate against the restored schema.
- [ ] `scripts/verify.ps1` passes.
- [ ] A new evolution event is recorded in `state/evolution/2026-07-25-loop-state-schema.evolution.json` documenting the promotion.

## Related
[...]

PR #867, opened 2026-07-29
fix(governance): collapse loop-state schemas to single source of truth (#852)

### LINKED ISSUE #852: Two loop-state schemas for one artifact class — collapse to a single source of truth
## The duplication

Two files describe the same artifact class, with different `$id`s and divergent shapes:

| File | `$id` | Enforced by | Referenced by |
|---|---|---|---|
| `schemas/loop-state.schema.json` | `loop-state.schema.json` | **`demerzel_kit.py:50`** (`SCHEMA_DIR = ROOT / "schemas"`) at every write, via `run_afk_cycle.py` → `kit.write_artifact(..., schema="loop-state")`. Regression-tested at `scripts/test_run_afk_cycle.py:219-239`. | `policies/autonomous-loop-policy.yaml:20` |
| `logic/loop-state.schema.json` | `https://github.com/GuitarAlchemist/Demerzel/schemas/loop-state` | **nothing** — no CI check validates `state/loops/*` against it. The only thing that ever read it was a manually-run `/demerzel audit`. | `policies/autonomous-loop-policy.yaml:564` |

`schemas/` is authoritative: it governs writes, it is regression-tested, and reality conforms to it. `logic/` is doctrine-referenced only.

## Why this is the root cause, not a tidiness complaint

The `logic/` copy sat untouched from 2026-03-15 while the system was built against the `schemas/` copy. By 2026-07-25 it had drifted to requiring 13 fields, five of which **no instance has ever carried and no script has ever read** — and its `target_repo` enum was `[ix, tars, ga]` while every real instance is `repo: demerzel`. It had become unsatisfiable: no emitter, however enriched, could have passed it.

That is what the audit in #839 tripped over. The two copies are now aligned at 7-vs-7 required fields with identical enums, but **nothing prevents them drifting apart again** — there is still no CI check binding them, and the next contributor to edit one has no signal that the other exists.

Alignment is a snapshot. Deduplication is the fix.

## Suggested resolution

1. Delete `logic/loop-state.schema.json`, or reduce it to a `$ref` at the `schemas/` copy so there is exactly one shape definition.
2. Repoint `policies/autonomous-loop-policy.yaml:564` at the surviving file.
3. Settle the `$id` — the two currently differ, so a `$ref` across them would not resolve as written.
4. Extend `validate_governance.py` to validate `state/loops/*.loop.json` against the surviving schema, so this is caught by CI rather than by a hand-run audit skill eleven months later.

Step 4 matters most. Without it the next divergence is found the same way this one was: by accident.

## Knock-on for #834

**#834's acceptance criteria name the wrong file.** They call for restoring the rich auditability fields to `logic/loop-state.schema.json`, but the file governing writes is `schemas/`. Enriching the unenforced copy would change nothing about what the emitter is allowed to produce.

Sequence should be: collapse to one source of truth here, *then* decide whether `worker`/`reviewer`/`convergence` are actually wanted in the surviving schema. #834 should be retargeted or closed-superseded accordingly.

## Related

- #839 — aligned the two copies to 7-vs-7 as a side effect; did not deduplicate
- #834 — targets the unenforced copy
- #641 — unratified proposal on not weakening guardrails; adjacent

spareilleux:
Closes #852.\n\n## What changed\n\n- Deleted the unenforced duplicate logic/loop-state.schema.json.\n- Retained schemas/loop-state.schema.json as the single source of truth (the copy enforced by demerzel_kit.py at write time).\n- Repointed all active references in policies/autonomous-loop-policy.yaml, .claude/skills/demerzel-audit/SKILL.md, .claude/skills/demerzel-loop/SKILL.md, and pipelines/governance-audit.ixql.\n- Extended scripts/validate_governance.py with a new check_loop_states() gate that validates every state/loops/*.loop.json instance against schemas/loop-state.schema.json in CI.\n- Updated state/beliefs/2026-07-25-schema-conformance.belief.json and state/evolution/2026-07-25-loop-state-schema.evolution.json to record the collapse.\n- Updated governance-health.json via uild_manifest.py.\n\n## Verification\n\n- python scripts/validate_governance.py → 17 evolution logs, 4 loop states, 129 compliance reports valid.\n- pwsh scripts/verify.ps1 → passes.\n- python scripts/build_manifest.py → 391 artifacts, 157 edges, green.\n\n## Notes\n\n- Issue #834 should be retargeted at schemas/loop-state.schema.json if the richer auditability fields are restored.\n- The surviving schema keeps $id: loop-state.schema.json because it is loaded by path from schemas/ and no URL $ref is required.


spareilleux (commented):
## Cross-Model Review

### Findings

- [CORRECTNESS] `governance-health.json` diff swaps `latest_artifact` from `afk-graduated-self-merge` to `contracts/galactic-protocol.md` and bumps `computed_at` by a week — this is unrelated to the schema-consolidation change and isn't explained anywhere in the diff. Either this belongs to a separate commit that got squashed in, or it's stale/unrelated state being carried along. Needs justification or should be split out.

- [CORRECTNESS] `check_loop_states` glob pattern is `*.loop.json`, but nothing in the diff confirms this matches the actual on-disk naming convention used elsewhere (other checks reference `state/loops/*.json` in `governance-audit.ixql`, without the `.loop.` infix). If the naming conventions diverge, this check silently validates zero files forever and reports false-green via the "nothing to validate" NOTE path.

- [CORRECTNESS] Draft7Validator is used to validate a schema whose `$schema` field declares `https://json-schema.org/draft/2020-12/schema`. Mismatched draft version can cause silent validation drift (e.g., differing `additionalProperties`/`prefixItems` semantics). Should use `jsonschema.validators.validator_for(schema)` for correctness rather than hardcoding Draft7.

- [CORRECTNESS] The `except ImportError` skip path returns `(0, 0)` — a missing dependency in CI causes this gate to silently pass with no failure signal and no visible warning in the final summary line (only a printed SKIP, easy to miss in CI logs). This defeats the purpose of adding the check as a governance gate.

- [ANTI-SLOP] The deleted `logic/loop-state.schema.json`'s description field is a wall of prose explaining the reconciliation rationale, changelog, and cross-references (issue numbers, "richer auditability fields," etc.) — this belongs in commit message / ADR, not embedded in a JSON schema description field that gets parsed and surfaced by tooling. This is being deleted here, which is good, but flags that whoever wrote it treats schema files as changelog/documentation dumping grounds. Worth a note if the surviving `schemas/loop-state.schema.json` has the same problem (not shown in diff, can't confirm).

- [LOLLI DETECTION] This diff is a clean consolidation (single source of truth for loop-state schema, all references updated, CI enforcement added) — this is the *opposite* of lolli-shaped: it removes a duplicate artifact rather than adding an orphaned one, and the new CI check (`check_loop_states`) has a named consumer (`main()`, wired into exit code). Good faith reduction of governance surface.

- [CONSTITUTIONAL COMPLIANCE — Article 3 Reversibility] Deleting the schema file is fully reversible via git; no state loss since the diff confirms no instance data ever used the removed fields (per the now-deleted description text). Acceptable.

- [CONSTITUTIONAL COMPLIANCE — Article 4 Proportionality] Adding a full new validation function + CI wiring for a single schema consolidation is proportionate — it directly prevents regression to the two-schema-drift problem being fixed. Not overbuilt.

### Verdict
**REQUEST_CHANGES** — unexplained/unrelated `governance-health.json` artifact swap needs justification, and the CI validator has two silent-failure paths (ImportError skip, draft-version mismatch, glob pattern assumption) that undermine the enforcement goal of this change.

---

PR #888, opened 2026-07-29
feat(afk): extract backend invocation paths into adapter modules (#873)

spareilleux:
Extracts the existing `--backend claude-code`, `--backend local`, and `--backend remote` invocation paths from `scripts/run_afk_cycle.py` into dedicated adapter modules under `scripts/afk_backends/`.

Closes #873.

## Changes

- `scripts/afk_backends/claude_code.py` — `ClaudeCodeBackend` (headless `claude -p` on the interactive subscription, strips `ANTHROPIC_API_KEY`).
- `scripts/afk_backends/sandcastle.py` — `SandcastleBackend` (Podman sandcastle via the sibling `../afk-harness` project).
- `scripts/afk_backends/remote.py` — `RemoteBackend` (no-op cloud-worker seam, to be implemented in #879).
- `scripts/run_afk_cycle.py` — governor now selects a backend via `get_backend(name)` and delegates `prepare()` and `invoke()` to the adapter. Removed the monolithic backend functions.
- `scripts/test_run_afk_cycle.py` — tests now mock `get_backend` and the adapter classes instead of internal helper functions.

## Verification

- `python -m unittest discover -s scripts -p "test_*.py"` → pass (357 tests).
- `python scripts/validate_governance.py` → pass (146 valid, 0 invalid).
- `python scripts/build_manifest.py` → green.
- `pwsh scripts/verify.ps1` → blocked by missing `tree-sitter` on PATH (pre-existing environment issue, not related to this change).

## Backward compatibility

No CLI or behavioral changes. The same `--backend` values work, the same `{branch, commits, blocked}` result shape is returned, and the budget gate still runs before every worker invocation.

Issue #873, opened 2026-07-29
[AFK-v0.2][AFK] Extract existing backends into scripts/afk_backends/ adapters


spareilleux:
Parent: #471\nBlocked by: #872\n\n## What to build\n\nMove the two existing AFK invocation paths out of scripts/run_afk_cycle.py into a new scripts/afk_backends/ package, without changing behavior. This is a pure refactoring slice that proves the adapter contract from #872 can carry the current implementation.\n\nThe existing paths are:\n- _invoke_harness_claude_code (headless claude -p subscription backend)\n- _invoke_harness (sandcastle / ../afk-harness local backend)\n- _invoke_harness_remote (unimplemented Vercel seam)\n\nEach should become a concrete adapter module:\n- scripts/afk_backends/claude_code.py\n- scripts/afk_backends/sandcastle.py\n- scripts/afk_backends/remote.py (stub, returning the current not-implemented blocked result)\n\n
un_afk_cycle.py should then select the backend by name and call it through the protocol, not by branching on the backend string.\n\n## Acceptance criteria\n\n- [ ] New package scripts/afk_backends/ exists with __init__.py (the protocol) and the three adapter modules.\n- [ ] 
un_afk_cycle.py no longer contains _invoke_harness, _invoke_harness_claude_code, or _invoke_harness_remote.\n- [ ] python -m unittest discover -s scripts -p 'test_*.py' still passes.\n- [ ] python scripts/validate_governance.py still passes.\n- [ ] pwsh scripts/verify.ps1 still passes.\n- [ ] No behavioral change: the same backends are selected by the same --backend argument and produce the same results.\n\n## Blocked by\n\n- #872 — the adapter contract must be defined first.

spareilleux:
🤖 **Demerzel Governance Notice**

This issue has been added to the [Demerzel Governance](https://github.com/orgs/GuitarAlchemist/projects/2) project board.

**Governance checklist:**
- [ ] Reviewed against [Asimov constitution](https://github.com/GuitarAlchemist/Demerzel/blob/master/constitutions/asimov.constitution.md) (Articles 0-5)
- [ ] Checked [default constitution](https://github.com/GuitarAlchemist/Demerzel/blob/master/constitutions/default.constitution.md) compliance (Articles 1-11)
- [ ] Risk assessed per [autonomous loop policy](https://github.com/GuitarAlchemist/Demerzel/blob/master/policies/autonomous-loop-policy.yaml)
[...]

Issue #872, opened 2026-07-29
[AFK-v0.2][HITL] Define the AFK backend adapter contract


spareilleux:
Parent: #471\n\n## What to build\n\nDefine the interface that every AFK implementation backend must satisfy, whether the tool runs on the desktop (Claude Code, GitHub Copilot CLI, JetBrains AI) or in the cloud (Vercel sandbox, GitHub Codespaces, remote worker).\n\nThe contract must be narrow enough to implement in a small adapter module, but complete enough that the governor (scripts/run_afk_cycle.py) can treat every backend identically. Today the governor hardcodes two invocation paths: claude -p and a Podman sandcastle harness. We want a single abstraction behind which any tool can plug in.\n\nThe contract should cover the lifecycle already implicit in the existing backends:\n\n1. **prepare** — check prerequisites (e.g., Podman running, tool installed, cloud endpoint reachable).\n2. **invoke** — run the tool against a fresh clone of the repo for one issue.\n3. **observe** — capture whether the tool produced a branch and commits.\n4. **collect** — return {branch, commits, blocked} in the same shape the governor already uses.\n5. **cancel** — abort an in-flight invocation if the cycle is halted.\n6. **estimate_cost** — return a rough cost estimate for the AIW budget gate.\n\n## Acceptance criteria\n\n- [ ] ADR or design note in docs/solutions/harness/ explaining the contract and the design choices.\n- [ ] Python protocol / abstract base class in scripts/afk_backends/__init__.py defining AFKBackend.\n- [ ] The protocol documents the exact return type and error contract for each method.\n- [ ] A draft mapping of existing backends (claude-code, local) to the new contract is included in the design note.\n- [ ] No implementation of concrete backends in this slice — design only.\n\n## Blocked by\n\nNone — can start immediately.\n\n## Related\n\n- #471 — Cherny-style agent loop and budget router integration\n- #863 — local backend misattributes metered spend\n- #776 — run_afk_cycle.py is not scheduled

spareilleux:
🤖 **Demerzel Governance Notice**

This issue has been added to the [Demerzel Governance](https://github.com/orgs/GuitarAlchemist/projects/2) project board.

**Governance checklist:**
- [ ] Reviewed against [Asimov constitution](https://github.com/GuitarAlchemist/Demerzel/blob/master/constitutions/asimov.constitution.md) (Articles 0-5)
[...]

PR #886, opened 2026-07-29
feat(afk): define backend adapter contract (#872)

spareilleux:
Closes #872.

## What

Defines the AFK backend adapter contract so the governor can treat any AI tool (desktop or cloud) identically.

## Changes

- `scripts/afk_backends/__init__.py` — new abstract `AFKBackend` class with methods:
  - `prepare()` — check prerequisites before a cycle starts
  - `needs_local_repo()` — declare whether the governor must prepare an ephemeral clone
  - `invoke(issue, repo_path)` — run the tool and return `{branch, commits, blocked}`
  - `estimate_cost(issue)` — rough cost estimate for the AIW budget gate
  - `cancel()` — best-effort abort (default no-op)
- `docs/solutions/harness/2026-07-29-afk-backend-adapter-contract.md` — design ADR explaining the contract, the registry plan, and the mapping to existing backends.
- `scripts/test_afk_backends.py` — contract tests verifying the ABC cannot be instantiated and a concrete implementation satisfies the protocol.

## Verification

- `python -m unittest scripts.test_afk_backends` → 2 tests pass
- `python scripts/validate_governance.py` → 146 valid, 0 invalid
- `pwsh scripts/verify.ps1` → passes
- `python scripts/build_manifest.py` → green

## Next steps

- #873 — extract existing backends into adapter modules
- #875 — add backend registry loader
- #877 — fix local backend spend attribution
- #879 — implement cloud-worker backend
- #882 — add shell adapter as proof of abstraction
- #884 — update docs and policy

PR #889, opened 2026-07-29
state(afk): record backend-adapter belief and evolution (#873)

spareilleux:
Adds the mandatory state maintenance for the AFK backend adapter extraction (PR #888).\n\n- state/beliefs/2026-07-29-afk-backend-adapter.belief.json records that the governor now uses a backend-adapter abstraction.\n- state/evolution/2026-07-29-afk-backend-adapter.evolution.json records the contract creation (#872) and backend extraction (#873) events.\n\nValidated with python scripts/validate_governance.py → 18 evolution logs valid, 0 invalid.

### DIFF
--- a/state/beliefs/2026-07-29-afk-backend-adapter.belief.json
+++ b/state/beliefs/2026-07-29-afk-backend-adapter.belief.json
@@ -1,0 +1,23 @@
+{
+  "proposition": "The Demerzel AFK implement governor uses a backend-adapter abstraction so that desktop, container, and cloud AI tools can be plugged in without changing the governor.",
+  "truth_value": "T",
+  "confidence": 0.92,
+  "evidence": {
+    "supporting": [
+      {
+        "source": "feat/afk-backend-adapter-contract-872 (#886)",
+        "claim": "Defined the AFKBackend abstract contract (prepare, needs_local_repo, invoke, estimate_cost, cancel) and documented the adapter architecture in docs/solutions/harness/2026-07-29-afk-backend-adapter-contract.md.",
+        "timestamp": "2026-07-29T00:00:00Z",
+        "reliability": 0.95
+      },
+      {
+        "source": "feat/extract-afk-backends-873 (#888)",
+        "claim": "Extracted the existing claude-code, local (sandcastle), and remote invocation paths from scripts/run_afk_cycle.py into scripts/afk_backends/claude_code.py, sandcastle.py, and remote.py. The governor now selects backends via get_backend() and delegates prepare()/invoke() to the adapter. All 357 unit tests pass.",
+        "timestamp": "2026-07-29T00:00:00Z",
+        "reliability": 0.98
+      }
+    ]
+  },
+  "last_updated": "2026-07-29T00:00:00Z",
+  "evaluated_by": "demerzel"
+}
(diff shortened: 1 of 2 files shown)


spareilleux (commented):
## Cross-Model Review

### Findings

- [CORRECTNESS] The evolution artifact's `metrics` block reports `citation_count: 0`, `compliance_rate: 0.0`, and both promotion/deprecation candidates as `false`, yet the assessment recommends `"investigate"` with 0.85 confidence — this is internally consistent (new artifact, no usage data yet), but the belief file asserts confidence 0.92 truth_value "T" for a capability with zero citations. Confidence appears to be based on test-pass evidence rather than actual governance usage. Not a hard error, but conflates "code exists and passes tests" with "capability is proven in production."

- [CORRECTNESS] No rollback/deprecation path is documented if the adapter abstraction proves wrong (e.g., if a future backend can't fit the `prepare/needs_local_repo/invoke/estimate_cost/cancel` contract). This is a state-tracking diff, not the code diff, so the actual PRs (#886, #888) should be reviewed separately for test coverage of `estimate_cost`/`cancel` — those aren't mentioned as exercised in the "357 unit tests pass" claim.

- [LOLLI DETECTION] Two governance artifacts (belief + evolution record) are created for a single internal refactor (extracting existing invocation logic into adapter modules). No named consumer of these records is specified — who queries `state/evolution/*.json` and acts on the `"recommendation": "investigate"` field? If no downstream process consumes this, it's governance weight added without a defined feedback loop closing the value.

- [LOLLI DETECTION] `citation_count: 0` and `promotion_candidate: false` at creation time is honest, but raises the question: why generate the evolution artifact *before* any citation data exists? If the artifact is speculative bookkeeping, this is process overhead in advance of evidence — borderline against Article 4 (Proportionality), since two structured JSON files are the governance cost for an internal code reorganization with no behavior change claimed.

- [CONSTITUTIONAL COMPLIANCE] Article 3 (Reversibility): the evolution log states old functions (`_invoke_harness`, `_invoke_harness_claude_code`, `_invoke_harness_remote`, `_ensure_podman`) were *removed*, not deprecated-in-place. No mention of a rollback plan or feature flag if the new adapter path regresses one of the three backend types silently. Given this is the state-tracking record for a structural refactor of a production governance script, absence of a stated rollback path is a gap.

- [CONSTITUTIONAL COMPLIANCE] Article 9 (Bounded Autonomy): the belief record is self-evaluated (`"evaluated_by": "demerzel"`) with confidence 0.92 — the system is grading its own architectural change. No external/independent verification step is referenced for a structural change to the AFK governor itself (the component that governs autonomous behavior). Self-assessed high confidence on a change to the autonomy-governing script is the kind of self-referential grading Article 9 should be scrutinizing.

- [ANTI-SLOP] Minor: "capability" as `artifact_type` and generic `context` prose in `events` is acceptable, but the two records largely duplicate the same narrative (contract definition, then extraction) across both files — could be one record with clearer phase tracking instead of parallel belief + evolution duplication.

### Verdict

REQUEST_CHANGES — self-evaluated confidence on a change to the autonomy-governing script (Article 9), plus no rollback path for removed legacy functions (Article 3); clarify who consumes the evolution artifact before merging.

---

Issue #882, opened 2026-07-29
[AFK-v0.2][AFK] Add a second desktop adapter to prove tool-agnostic abstraction


spareilleux:
Parent: #471
Blocked by: #875

## What to build

Add a second desktop backend adapter to prove the abstraction from #872 works for a non-Claude tool. The simplest proof is a generic shell-out adapter that invokes any CLI tool that follows a documented contract.

The new backend `shell` runs an arbitrary command from `config/afk-backends.yaml`:

```yaml
backends:
  shell:
    adapter: scripts.afk_backends.shell.ShellBackend
    execution_mode: desktop
    provider: shell-local
    config:
      command: ["/path/to/my-ai-tool", "implement", "--repo", "{repo}", "--issue", "{issue_number}"]
      timeout: 1800
```

The adapter interpolates `{repo}`, `{issue_number}`, `{title}`, `{body}`, and `{branch}` into the command, runs it in the ephemeral clone, and parses the same `{branch, commits, blocked}` JSON from the last line of stdout (matching the sandcastle harness convention).

This is not meant for production use with untrusted commands; it is an escape hatch and a demonstration that the backend wall is real.

## Acceptance criteria

- [ ] `scripts/afk_backends/shell.py` implements the adapter contract.
- [ ] `config/afk-backends.yaml` contains a commented-out example `shell` backend.
- [ ] The adapter validates that the configured command is a list of strings and refuses shell metacharacters.
- [ ] Unit tests cover: successful run, command failure, JSON parse failure, missing config, timeout.
- [ ] `python scripts/validate_governance.py` and `pwsh scripts/verify.ps1` pass.

## Blocked by

- #875 — registry loader must exist before the new adapter can be registered and loaded.

## Related

- #872 — adapter contract definition
- #873 — backend extraction refactoring

spareilleux:
🤖 **Demerzel Governance Notice**

This issue has been added to the [Demerzel Governance](https://github.com/orgs/GuitarAlchemist/projects/2) project board.

**Governance checklist:**
- [ ] Reviewed against [Asimov constitution](https://github.com/GuitarAlchemist/Demerzel/blob/master/constitutions/asimov.constitution.md) (Articles 0-5)
- [ ] Checked [default constitution](https://github.com/GuitarAlchemist/Demerzel/blob/master/constitutions/default.constitution.md) compliance (Articles 1-11)
[...]

Issue #890, opened 2026-07-29
Adopt BAML for strongly-typed prompts and schema verification


spareilleux:
Integrate BAML (Boundary AI Markup Language) to define prompts as strongly-typed functions and validate LLM outputs. Update verify.ps1 to generate clients automatically.

spareilleux:
🤖 **Demerzel Governance Notice**

This issue has been added to the [Demerzel Governance](https://github.com/orgs/GuitarAlchemist/projects/2) project board.

**Governance checklist:**
- [ ] Reviewed against [Asimov constitution](https://github.com/GuitarAlchemist/Demerzel/blob/master/constitutions/asimov.constitution.md) (Articles 0-5)
- [ ] Checked [default constitution](https://github.com/GuitarAlchemist/Demerzel/blob/master/constitutions/default.constitution.md) compliance (Articles 1-11)
- [ ] Risk assessed per [autonomous loop policy](https://github.com/GuitarAlchemist/Demerzel/blob/master/policies/autonomous-loop-policy.yaml)
- [ ] [Reconnaissance](https://github.com/GuitarAlchemist/Demerzel/blob/master/policies/reconnaissance-policy.yaml) completed if cross-repo

*Governed by Demerzel under the [governance mandate](https://github.com/GuitarAlchemist/Demerzel/blob/master/constitutions/demerzel-mandate.md).*

spareilleux:
Reconciled during the GA Cloud Factory/local-work inventory. The concrete Demerzel adoption work is already represented by merged PRs #908 (tracked Python client + typed DSP boundary), #921 (consumer-facing clients moved back to their owning repos), #922 (subscription-backed typed grader with fail-closed/tool-free transport), and #924 (CI drift enforcement). The isolated demerzel-bot comparison later recorded a do-not-expand verdict: BAML may remain a typed parser at selected leaf boundaries, followed by independent canonical/domain validation; it is not a fleet schema, router, budget, authorization, or merge authority. Closing this broad issue prevents the remaining agent-implement label from dispatching duplicate adoption work. Any future expansion should be a new narrow issue with a named consumer and measurable gain.

PR #908, opened 2026-07-29
feat(baml): three tracked clients + single-sourced DSP safety bounds (ADR-0005)

spareilleux:
Completes the BAML adoption that ADR-0005 declared but left partly unbuilt, closes the Article 4/7/9 gaps in the gate that consumes it, and **recovers work that had been lost from the shared checkout**.

## Why this branch exists

The entire BAML adoption was wiped from the `Demerzel` working tree mid-session. The reflog:

```
HEAD@{2}: checkout: moving from feat/second-desktop-adapter-882 to master
HEAD@{1}: reset: moving to origin/master        <- hard reset
HEAD@{0}: checkout: moving from master to feat/second-desktop-adapter-882
```

A concurrent session reset to `origin/master` and switched branches, having first run `git stash -u`. Everything survived in `stash@{0} "Unrelated work before #882"` and was recovered into an isolated worktree rather than back into the shared checkout, which is still mid-flight on #882. The stash was **applied, not popped**, so it remains as a backup.

## Client generation (ADR-0005 §Decision.3)

`baml_project.baml` declared only `python/pydantic`; the ADR promises `typescript` (chatbots) and `rust` (ix, hari). Both now generate.

- Each generator needs a **distinct `output_dir`** — BAML writes `<output_dir>/baml_client`, so two generators sharing a directory silently clobber each other.
- Python stays at the repo root so `from baml_client import b` resolves per §Consequences.
- The Rust output is a **module tree** (`mod.rs` + `types/` + `functions/`), not a crate — no generated `Cargo.toml`, so the consuming crate owns the runtime dep.

All three trees are **committed**: sibling repos copy or reference them and cannot run `baml generate` from their own build. `verify.ps1` regenerates and fails on drift — the same committed-and-CI-diffed discipline as `governance-manifest.json`. This also restores the `verify.ps1` BAML step §Decision.4 requires, which had likewise been lost.

## Threshold duplication

`0.15` (THD) and `0.85` (StdDev) were stated **twice** — Python literals in `run_hari_consensus()` *and* English prose inside the `EvaluateSignalSwarm` prompt. Retuning one would silently desynchronise the code gate from the model gate.

Both now read `logic/dsp-safety-bounds.yaml`; the prompt interpolates `{{ bounds.thd_max }}` through a new `SafetyBounds` input. ADR-0002, harvest don't declare.

## `validate_dsp_loop.py`

**Fail-open (the one that matters).** `best_safe` was seeded from the cache before the loop, so a cycle-1 compile error exited **0** printing `CONSTITUTIONAL RULES SATISFIED` for a value the run never validated. Only a passing cycle in *this* run may set it now; an aborted run exits 1 and proposes nothing.

**`--frequency` was unusable off its 5.0 default.** The fundamental's DFT bin was hardcoded `4 <= k <= 6`, so at any other frequency the fundamental landed in the high-frequency band, THD exploded, and no value could ever pass. The bin is now derived from frequency and sample rate, with a guard for unusable ranges.

**Article 9.** A cached parameter reshaping the next run's search space is autonomous enactment. The warm start is opt-in (`--use-cached-bounds`) and **re-validates** the carried-forward value on cycle 1 rather than assuming it — which also surfaces a stale cache instead of silently capping the search ceiling.

**Article 7.** Each run writes `state/dsp-validation/` carrying `recursion_depth`, `article_4_check`, `requires_authorization`, bounds provenance and per-cycle votes. `cached_value_seen` is recorded whether or not it was enacted, so the boundary is auditable rather than implicit.

Votes now carry `self_trust`, which `SwarmVote` requires and the old dicts omitted.

## Three graders, one bounds file

A concurrent session had independently rewritten this script to add `HariSession` (live stdio JSONL to a compiled `hari-core serve`). That work is **preserved and merged**, not clobbered:

| Grader | Status |
|---|---|
| `hari-live` | verified — substrate returned `signal_quality: True` |
| `deterministic` | verified — default fallback |
| `--use-baml` | wired and type-correct, **never made a live call** |

All three converge on **distortion 0.5078**.

## Verified

- 3 clients generate; drift check passes
- cold run exits 0 at 0.5078
- seeded cache + broken generator exits **1**
- `--frequency 20` converges to the same 0.5078 (correct — the `tanh` shaper is frequency-independent)
- stale cache detected and discarded
- recovered `test_recursive_learning_eval_schema.py`: **13 tests passing** under the CI invocation

**Not verified:** `--use-baml` end to end. `baml-py` is not installed and no provider key is present.

## Second commit

`2fff306` carries the concurrent session's recovered work verbatim, kept separate so it stays attributable: the recursion-depth/Article-4 schema, its 13 tests, the `implementation_status` block declaring the eval cycle has no executor, `monitor_baml_and_learning.py`, and the `aiw_prompt_pack.py` BAML migration.

Known gap, not fixed here: `monitor_baml_and_learning.py` counts only `baml_client/*.py`, so it under-reports now that the TS and Rust clients exist, and it reads the `.tmp/` cache rather than the audited records.


spareilleux (commented):
## Cross-Model Review

### Findings
- **CORRECTNESS**: No logic errors or edge case issues detected. The code is auto-generated, and functions appear well-structured with appropriate parameter handling.
- **SECURITY (OWASP)**: No injection risks or secrets exposure. The handling of parameters appears secure, and there is no apparent usage of insecure defaults.
- **ANTI-SLOP**: The code is auto-generated, so certain variable and class names (e.g., `DoNotUseDirectlyCallManager`) appear unnecessarily complex and could potentially confuse developers.
- **LOLLI DETECTION**: No creation of unconsumed artifacts detected. All newly introduced elements seem to integrate well with existing infrastructure and have expected usage paths.
  - **Article 3 (Reversibility)**: The changes introduce components that can be reverted by re-generating the code base using specified tools without changing core governance data.
  - **Article 4 (Proportionality)**: The newly added clients and configuration seem proportional to the functionality they aim to provide.
  - **Article 9 (Bounded Autonomy)**: The use of BAML and its abilities seems consistent with encouraging autonomy while respecting governance structures, in line with Article 9 guidelines.

### Verdict
APPROVE with confirmation that the code is well-structured, secure, and complies with constitutional requirements.

---

spareilleux (commented):
## Cross-Model Review

### Findings
- **CORRECTNESS**: No logic errors, off-by-one errors, or missing edge cases identified in this diff.
- **SECURITY (OWASP)**: There are no direct injection risks or exposure of secrets. The BAML adoption provides a strongly-typed DSL, which inherently can reduce certain types of injection risks, but the security perspective should remain vigilant as new features integrate with external client interactions.
- **ANTI-SLOP**: The naming convention appears clear, and classes such as `BamlAsyncClient` and `BamlStreamClient` are logically consistent. However, the use of `DO_NOT_USE_DIRECTLY_UNLESS_YOU_KNOW_WHAT_YOURE_DOING_RUNTIME` can be considered as a somewhat vague way to handle critical components.
- **LOLLI DETECTION**: The changes seem to provide a structured pathway for BAML's integration, which is actively used by consumers (Python, TypeScript, Rust clients). It does not appear to add governance weight unjustifiably.
- **CONSTITUTIONAL COMPLIANCE**: The modifications adhere to Article 3 (Reversibility) by ensuring that the BAML framework is integrated in a way that changes are trackable and reversible. Article 4 (Proportionality) is maintained, as the integration of BAML does not appear excessive alongside existing structures. Article 9 (Bounded Autonomy) is respected as the system requires explicit authorization for enacting parameters.

### Verdict
APPROVE with a confirmation that the integration of BAML seems procedurally sound and constitutionally aligned.

---

spareilleux:
Rebased onto `master` (now `d1208a8`) — was `DIRTY` against two files.

## Conflicts resolved

**`policies/recursive-learning-eval-policy.yaml`** — the entire conflict was **one comma** (`consumers"), and` vs `consumers") and`). Kept master's. Useful side effect: that was the branch's only change under a blocked path, so **this PR now touches zero blocked paths** and `risk-report` evaluates on merits rather than failing closed.

**`CONTEXT.md`** — took this branch's side, since the paths do become real here. Two corrections while resolving:

1. **Article 9 conflation.** The branch text said enacting a validated parameter as a control input *via* `--use-cached-bounds` "is an explicit authorization". The script itself disagrees — it prints *"This is a PROPOSED parameter (understanding). Enacting it as a control input requires human authorization"*, and `--use-cached-bounds` only authorizes a **warm start** (narrow the search space to a prior value, which cycle 1 then re-validates). CONTEXT.md is the authority document, so reworded to keep those two acts distinct.
2. **Superseded assessment.** `docs/research/2026-07-28-baml-adoption-assessment.md` is committed and recommends **defer**. Landing this without a pointer leaves a reader who finds it believing BAML is off the production path. Added an explicit "ADR-0005 supersedes the defer recommendation, adoption tracked in #890" clause rather than silently contradicting it.

Also recorded the fail-open fix in the glossary entry — `validated_this_run` can only be set by a cycle that passes in *this* run, so an aborted or non-converging run exits non-zero instead of re-proposing a cached value. That was a real gate failure and it is genuinely closed.

## Verified

- 417 tests pass; `scripts/validate_governance.py` → `PASSED: all checks green`
- `validate_dsp_loop.py` and `monitor_baml_and_learning.py` agree on `.tmp/dsp_validation/parameter_cache.json` — no born-dead path mismatch
- No blocked paths (confirmed against `agent-blackbox.policy.json`)

## Not merging — three findings for the owner

**1. Generated clients sit outside the runtime-code exception.** `CLAUDE.md:9` says *"Demerzel contains no runtime code — only governance artifacts. (Exception: `scripts/` and `.github/`.)"* This PR adds `baml_client/`, `clients/rust/`, and `clients/typescript/` at the repo root, which is neither. **#859** is the open PR that relaxes exactly this rule — it should land first, or this PR violates the principle at merge time. Ordering matters.

**2. 747 new lines of Python, zero tests.** The suite is 417 before and after. `validate_dsp_loop.py` is 641 lines and is described as a *safety gate*; it already shipped one fail-open defect. A gate with no test asserting it fires is the born-dead pattern.

**3. No consumer.** Nothing schedules or invokes either script — no workflow, no cron, no caller. `monitor_baml_and_learning.py` is referenced by literally nothing in the tree. Per the anti-lolli rule this is a producer without a consumer, and per the silent-loop-death pattern a monitor with no scheduler is dead on arrival.

None of the three is a defect in the code as written — they are gaps between it and the repo's own rules. Your call on sequencing; the branch is clean and mergeable whenever you want it.

PR #911, opened 2026-07-29
feat(afk): add Junie CLI backend adapter

spareilleux:
## What

Adds a JetBrains Junie CLI backend adapter to the AFK backend registry, continuing the tool-agnostic AFK refactor.

## Changes

- `scripts/afk_backends/junie.py` — `JunieBackend` implementation:
  - `prepare()` checks that `junie` is on `PATH` and that any configured `api_key_env` is set.
  - `needs_local_repo()` returns `True` so the governor prepares an ephemeral clone.
  - `invoke()` builds a task prompt from the issue, runs `junie --headless <prompt>` in the repo path, and reads the resulting git state to detect the branch and commits created by the agent.
  - Returns the standard AFK result contract: `branch`, `commits`, `blocked`.
  - Configurable via `config/afk-backends.yaml`: `command`, `task_template`, `timeout`, `api_key_env`, `branch_prefix`.
- `config/afk-backends.yaml` — adds the `junie` backend entry, mapped to provider `junie-cli`, disabled by default until Junie is installed and configured.
- `scripts/test_junie_backend.py` — mocked unit tests covering:
  - local-repo requirement
  - missing command / missing tool / missing API key in `prepare()`
  - successful invocation that detects a new branch and commits
  - Junie nonzero exit returning a `blocked` message
- `README.md` + `governance-manifest.json` — sync schema and contract-schema counts to 49 + 11 (the recursive-learning-eval schema and two contract schemas were already present in the tree; the counts were stale).

## Verification

- `python -m unittest discover -s scripts -p "test_*.py"` → 411 pass
- `python scripts/build_manifest.py` → green
- `python scripts/validate_governance.py` → 18 evolution logs valid, 0 invalid

> `pwsh scripts/verify.ps1` still fails locally because `tree-sitter` is not installed; this is an environment gap, not a regression from this change.

Issue #902, opened 2026-07-29
A conflicted blocked-path PR silently loses risk-report and renders all-green


spareilleux:
Observed live on #887 today.

## What happened

#887 touches `schemas/` — a blocked path — so `Agent Blackbox / risk-report` correctly failed closed on the first push. Master then advanced (#872, #873, #875, #877), and the branch went to `mergeStateStatus: DIRTY`.

From that moment, **no `pull_request` workflow ran on the PR at all**. GitHub builds `pull_request` events against the merge ref (`refs/pull/N/merge`); when the merge is conflicted that ref cannot be computed, so no event is dispatched. `push`-triggered workflows kept running normally, which is what makes this hard to notice.

The result: the PR's status rollup showed **8/8 SUCCESS with `risk-report` entirely absent.** Not red. Not pending. Gone. A reviewer glancing at the checks would see a fully green blocked-path PR.

`reopened` did not help either — I closed and reopened the PR and still got no runs, because the merge ref was still unresolvable. Only rebasing onto master restored dispatch:

```
DIRTY    → push runs only, rollup 8/8 green, no risk-report
UNSTABLE → pull_request runs dispatch again (Agent Blackbox, Cross-Model Review, Governance Validate)
```

## Why this matters

`risk-report` is the fail-closed gate that requires human attestation for `.github/workflows/**`, `constitutions/**`, `policies/**`, `schemas/**`, `logic/*.schema.json`. Its whole value is that it is *present and red* on those PRs.

A conflict silently converts "blocked pending human review" into "no opinion", and the UI renders no-opinion as green. The gate did not fail — it stopped being asked. That is this repo's recurring failure mode: a guard that is alive and correct, pointed at nothing.

Branch protection would catch this **if** `risk-report` is a required check, since a required check that never reports blocks merge. Worth confirming that is actually configured, because if it is not, the only thing standing between a conflicted blocked-path PR and a green merge button is whoever is looking at it.

## Suggested fixes

1. **Confirm `risk-report` is a required status check** on `master`. This is the real defence; everything else is a nicety.
2. Consider a scheduled or `push`-triggered sweep that flags open PRs where `mergeStateStatus` is `DIRTY`/`UNKNOWN` **and** the diff touches a blocked path — the one case where absence of a check is dangerous rather than merely untidy.
[...]

Issue #896, opened 2026-07-29
AFK: approved `local` runs leak their budget reservation — 4 leaks wedge every lane


spareilleux:
Found by adversarial review of #887 (and independently while building it). **Not** introduced by that PR in the sense of being a new hole — it is a consequence of #863 being fixed correctly. `local` genuinely is metered, and metered jobs genuinely do need a receipt.

## The defect

`aiw_budget_gate.release()` refuses to reconcile a metered job without a trusted provider receipt:

```python
# scripts/aiw_budget_gate.py:347-350
if reservation.get("metered"):
    if receipt is None:
        raise ValueError("a trusted provider receipt is required to release a metered job")
```

The governor never passes one. `_budget_release` (`scripts/run_afk_cycle.py:154-173`, called at `:700`) calls `budget.release(...)` with no `receipt` argument, ever.

Before #887, `local` mapped to `codex-cli` (local-seat, `metered: False`), so release reconciled cleanly. Now that `local` is correctly `anthropic-api` (metered-cloud), **every approved `local` episode ends in a failed release and a permanently open reservation.**

## Why it escalates beyond the one lane

- The cycle ledger (`.octo/aiw-cycle-ledger.json`) is never reset anywhere in the repo.
- The parallel-packet cap is **provider-agnostic** (`aiw_budget_gate.py:436-437`).

So after 4 unreconciled leaks, `cycle.max_parallel_packets` is exhausted and the gate blocks **everything** — including the healthy default `claude-code` subscription lane, which has nothing to do with the leak. Confirmed empirically: `claude-code` reserve → `block: parallel_packet_cap_exceeded`.

Recovery today requires a manual CLI release with a receipt artifact.

## Not currently reachable

`local` is blocked from the governor unconditionally (`_budget_reserve` passes `approval=None`), and **nothing in the repo invokes `--backend local`** — verified across `.github/workflows/` and `scripts/`. Nothing schedules the governor at all (#776). So this is latent until someone deliberately drives an approved `local` run.

It is also **not silent**: each leak writes a stderr line and a `release_failures` ledger note — the machinery #794 added for exactly this.

## Options

1. **Plumb the receipt.** The harness would have to emit a signed/attributable receipt from `api.anthropic.com` that `_validate_receipt` accepts. Real work; correct end state.
2. **Runbook it.** Declare `local` reserve-only and document CLI-based release as the operator's job. Cheap, honest, keeps the leak visible.
[...]

Issue #879, opened 2026-07-29
[AFK-v0.2][AFK] Implement cloud-worker backend adapter


spareilleux:
Parent: #471\nBlocked by: #875\n\n## What to build\n\nImplement the --backend remote seam that is currently a stub in 
un_afk_cycle.py. The cloud worker adapter delegates to a remote HTTP endpoint that runs the agent in an isolated environment (Vercel, GitHub Codespaces, or a generic worker).\n\nThe adapter should follow the same contract as the desktop adapters from #873. The endpoint contract is:\n\n- **POST** <endpoint>/run with JSON body {repo_url, issue_number, title, body, branch_prefix, allowed_paths, max_iterations}.\n- **Response** {branch, commits, blocked, cost_usd} or a streaming status URL.\n- **GET** <endpoint>/status/<job_id> for polling if async.\n\nThe adapter should be configurable via config/afk-backends.yaml:\n\n`yaml\nbackends:\n  remote:\n    adapter: scripts.afk_backends.remote.RemoteBackend\n    execution_mode: cloud\n    provider: cloud-worker\n    config:\n      endpoint: https://example.com/afk-worker\n      timeout: 1800\n`\n\nBy default the adapter returns locked: cloud worker not configured so the backend is safe to register even when no endpoint is set.\n\n## Acceptance criteria\n\n- [ ] scripts/afk_backends/remote.py implements the full adapter contract.\n- [ ] The adapter reads its endpoint and timeout from the registry config.\n- [ ] The adapter handles HTTP errors, timeouts, and malformed responses gracefully.\n- [ ] Dry-run mode prints the endpoint and payload without sending a request.\n- [ ] Unit tests cover: successful response, blocked response, HTTP error, timeout, missing config.\n- [ ] python scripts/validate_governance.py and pwsh scripts/verify.ps1 pass.\n\n## Blocked by\n\n- #875 — the registry loader must exist before the adapter can read cloud config.\n\n## Related\n\n- #471 — Cherny-style agent loop and budget router integration\n- #776 — run_afk_cycle.py is not scheduled

spareilleux:
🤖 **Demerzel Governance Notice**

This issue has been added to the [Demerzel Governance](https://github.com/orgs/GuitarAlchemist/projects/2) project board.

**Governance checklist:**
[...]

PR #914, opened 2026-07-29
fix(governance): declare all registry providers, unwedge metered release, correct three docs

spareilleux:
Sweep of the governance issues found today. Four independent fixes; each verified by mutation rather than by reading the diff.

## 1. #896 — approved `local` runs leaked their reservation

`release()` demands a trusted provider receipt for a metered job (`aiw_budget_gate.py:347-350`) and the governor has none to offer, so a metered episode left its reservation open forever. `active_packets` never fell, and after `cycle.max_parallel_packets` such episodes the gate blocked **every** provider — including the free subscription lane that had nothing to do with it.

Adds `aiw_budget_gate.abandon()`: frees the slot, charges the **reserved estimate** to actual spend, and records the job under `unreconciled` with `receipt_verified: false`.

**It deliberately does not synthesise a receipt.** `_validate_receipt` checks structure, not authenticity — every field it requires is derivable from the reservation the governor just created — so a self-issued receipt would pass and reconcile metered spend at whatever we claimed, forever, self-certified. That restores liveness by destroying the control the receipt exists to provide. Charging the estimate is pessimistic on purpose: **unverified is not the same as free.**

## 2. Undeclared providers

`generic-shell` (#882) and `junie-cli` (#911) named providers no policy declared. Both failed closed, so they were safe but unrunnable — and the misconfiguration would only have surfaced as a mystery block the first time someone set `enabled: true`.

| provider | tier | approval | rationale |
|---|---|---|---|
| `generic-shell` | `local-seat` | no | runs a local command, no metered API |
| `junie-cli` | `metered-cloud` | **yes** | JetBrains AI bills a quota; its own registry config references `JUNIE_API_KEY` |

`junie-cli` is pinned by id in the schema for the same reason `anthropic-api` is: the tier-keyed guards constrain shape *given* a tier, and tier is editable in the same diff that would downgrade it.

The registry↔policy test is now **unconditional**. Scoping it to `enabled` entries is exactly what let both of these through; checking every entry is what catches the next one.

## 3. `remote` — attribution marked provisional, not fixed

What `remote` bills depends entirely on the endpoint an operator configures, so `claude-code-cli` is a false statement — a remote HTTP worker is not a local subscription seat. But choosing its tier is #879's call, not this PR's. It is `enabled: false`, so the risk is latent; the config now says so explicitly and warns to re-attribute before enabling. Filed separately.

## 4. Three documents describing things that do not exist

**`CONTEXT.md`** — the file `CLAUDE.md` names as *the authority* — described **seven** paths, none of which are in the repository: `baml_src/`, `baml_client/`, `clients/`, `scripts/validate_dsp_loop.py`, `logic/dsp-safety-bounds.yaml`, `state/dsp-validation/`, and ADR-0005 — including a markdown link to the missing bounds file. Both glossary entries now say **PROPOSED, NOT IN TREE** and point at the committed decision to defer BAML (`docs/research/2026-07-28-baml-adoption-assessment.md`, #890). It also records that the local `validate_dsp_loop` prototype **fails open**: with a warm cache it exits **0** printing `CONSTITUTIONAL RULES SATISFIED` on a compilation error. I ran it.

**`state/learning/README.md`** cited a skill that exists nowhere (`.agent/skills/demerzel-meta-brainstorm/SKILL.md`) and the epistemic constitution at a path from a different repo's layout. Rewritten to reconcile the two competing specifications for that directory, state that **neither has ever produced a file**, and note that its behavioral test asserts a result *"stored in state/learning/meta-metrics.json"* — a file that has never existed. Behavioral tests are prose counted by `governance-manifest.json`, not executed, so it passes by existing.

## Verification

| mutation | result |
|---|---|
| disable `abandon` (restore the #896 leak) | **2 failures** |
| `abandon` charges `$0` instead of the estimate | **1 failure** |
| unmutated | 417 pass |

```
claude-code  claude-code-cli  local-seat     allow  []
local        anthropic-api    metered-cloud  block  ['provider_requires_manual_approval']
remote       claude-code-cli  local-seat     allow  []
shell        generic-shell    local-seat     allow  []
junie        junie-cli        metered-cloud  block  ['provider_requires_manual_approval']
```

`validate_governance` PASS · `belief_lint` 0 FAIL.

## Review

Touches `schemas/` — **blocked path**. `risk-report` will fail closed. Not applying `agent-blackbox-reviewed`; that is yours.

Note that the blocked-path gate currently has no teeth regardless — `master` has no required status checks, which is #902. I attempted a ruleset for that today and it is **left disabled**; see the issue for why.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -48,21 +48,25 @@ humanity) overrides everything. Constitutions are **append-only**.
 - **LOLLI / Resilience score (R)** — `R = injections_caught / injections_total`; the
   chaos-engineering metric for how well governance detects injected poison (dead
   bindings, orphaned branches, BS descriptions, unconsumed artifacts).
-- **BAML (Boundary AI Markup Language)** — the DSL (`baml_src/`) that declares prompts as
-  strongly-typed functions, enforcing LLM output schemas *in flight* rather than validating
-  them at rest (ADR-0005). One `.baml` source generates three clients: Python/Pydantic at
-  the repo root (`baml_client/`, so `from baml_client import b` resolves) plus TypeScript
-  and Rust under `clients/`, one per consumer language. Each generator needs a distinct
-  `output_dir` — BAML writes `<output_dir>/baml_client`, so two generators sharing a
-  directory silently clobber each other. The Rust output is a module tree, not a crate.
-- **validate_dsp_loop** — the self-correcting parameter gate (`scripts/validate_dsp_loop.py`):
-  binary-searches a DSP distortion parameter until a hexavalent swarm consensus clears the
-  bounds in [`logic/dsp-safety-bounds.yaml`](logic/dsp-safety-bounds.yaml). Two interchangeable
-  graders — deterministic threshold comparisons (default) and the typed BAML
-  `EvaluateSignalSwarm` function (`--use-baml`) — read the *same* bounds file, so the code
-  gate and the model gate cannot drift apart. Emits an audited record per run to
-  `state/dsp-validation/`; a validated parameter is a **proposal**, and enacting it as a
-  control input (`--use-cached-bounds`) is an explicit authorization, per Article 9.
+- **BAML (Boundary AI Markup Language)** — **PROPOSED, NOT IN TREE.** A DSL that would
+  declare prompts as strongly-typed functions, enforcing LLM output schemas *in flight*
+  rather than validating them at rest. The committed decision is to **defer**: see
+  [`docs/research/2026-07-28-baml-adoption-assessment.md`](docs/research/2026-07-28-baml-adoption-assessment.md)
+  ("do not put BAML on a production path or make it a cross-repo contract yet") and #890.
+  No `baml_src/`, `baml_client/`, or `clients/` exists in this repository, and there is no
+  committed ADR-0005. Do not write code against those paths. If adoption proceeds, note
+  that each generator needs a distinct `output_dir` — BAML writes `<output_dir>/baml_client`,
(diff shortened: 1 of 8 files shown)


spareilleux (commented):
## Cross-Model Review

### Findings

- **CORRECTNESS**: The addition of the `abandon` function is consistent with the logic to handle unreconciled reservations by permanently recording them. This appears to correctly address an edge case where reconciliation cannot happen due to lack of receipt verification.
- **SECURITY (OWASP)**: No immediate security risks observed, but care should be taken that the `reason` string is properly sanitized to avoid any potential injection risks when recorded or logged. Handling paths and file content seems properly contained.
- **ANTI-SLOP**: The use of terms like `junie-cli`, `unreconciled`, and `trusted_receipt_issuer` is contextually accurate and does not suffer from vague variable naming. The logic of handling metered and non-metered backends is clear and not unnecessarily complex.
- **LOLLI DETECTION**: The removal and adjustment to clarify non-existent code paths indicate good governance hygiene; it prevents unnecessary governance weight by not building around premature or unadopted systems.
- **CONSTITUTIONAL COMPLIANCE**: The changes are compliant with Article 9 (Bounded Autonomy) as it prevents autonomous actions without proper checks by ensuring that the mechanism to abandon a job does not bypass financial control measures. The changes support Article 3 (Reversibility) by allowing accurate tracking of unreconciled spend, ensuring financial transparency and reversibility of fiscal error.

### Verdict
APPROVE with confirmation that the changes bring clarity and accurate recording of billing processes.

---

PR #922, opened 2026-07-30
feat(baml): run the typed grader on the subscription, not a metered API

spareilleux:
Answers "can we avoid pumping on the API key?" — yes, and the fix turned out to be two things, one architectural and one that had been quietly true the whole time.

## The defect

`validate_dsp_loop.py --use-baml` called `b.EvaluateSignalSwarm(...)` directly. That sends to whatever client `baml_src` declares, and the declared client was:

```baml
client "openai/gpt-4o" // Default client config
```

`openai` appears **nowhere** in `state/driver/aiw-budget-policy.json`. The declared providers are five `local-seat` and five `metered-cloud` ids; OpenAI is not among them. So the repo's one model-based governance gate would have billed a provider the AIW budget gate has never heard of — the #863 shape exactly: a correct gate, pointed at a provider nobody is using. The `// Default client config` comment says nobody chose it.

## The architectural half

BAML's two useful halves don't need a provider. The generated client exposes them separately:

```python
req  = b.request.EvaluateSignalSwarm(...)   # render prompt — local, no socket
raw  = run_claude_code(prompt)              # completion — subscription, out of band
vote = b.parse.EvaluateSignalSwarm(raw)     # type + validate — local, no socket
```

The typed prompt and in-flight schema enforcement ADR-0005 was adopted for are untouched; only the wire changes. `claude-code-cli` **is** a declared provider — `local-seat`, `subscription-or-local`, no manual approval, first in `selection_order` — so the spend now lands inside a control that already knows how to classify it.

## The half that was quietly true

**Claude Code prefers `ANTHROPIC_API_KEY` over the claude.ai subscription when both are present**, and says so on stderr:

> ⚠ claude.ai connectors are disabled because ANTHROPIC_API_KEY or another auth source is set and takes precedence over your claude.ai login

Verified with the same prompt:

| | result |
|---|---|
| key set | `API Error: 400 You have reached your specified API usage limits` |
| key unset | a normal completion |

That is the actual pumping, and it explains the exhausted quota and the nine workflows failing on it. So `run_claude_code` strips `ANTHROPIC_API_KEY` (and `ANTHROPIC_AUTH_TOKEN`) from the **child** environment rather than trusting the parent's. A transport whose billing depends on an ambient variable is a coincidence, not a guarantee. `subscription_env()` is factored out so that property is assertable without spawning anything, and it copies rather than mutating `os.environ`.

## Fail-closed by construction

[...]

PR #923, opened 2026-07-30
fix(ci): stop the wiki sync dying on SIGPIPE (#917)

### LINKED ISSUE #917: wiki-cross-repo: sed|head SIGPIPE under pipefail fails the sync on every run
`Demerzel Cross-Repo Wiki Sync` has failed on **six consecutive runs** (2026-07-29 → 2026-07-30). Nothing tracks it, and no freshness guard flags it — the workflow is not stale, it is red and ignored.

## Root cause

`.github/workflows/demerzel-wiki-cross-repo.yml:96`

```bash
sed -n '1,/<!-- BEGIN DEMERZEL GOVERNANCE -->/p' repo/CLAUDE.md | head -50 >> wiki/Home.md
```

The step runs under `bash --noprofile --norc -e -o pipefail`.

`CLAUDE.md` up to that marker is longer than 50 lines, so `head -50` exits as soon as it has its quota and closes the pipe. `sed` is still writing, takes `EPIPE`, and exits non-zero:

```
sed: couldn't flush stdout: Broken pipe
##[error]Process completed with exit code 4
```

`pipefail` promotes that to a pipeline failure and `-e` aborts the step. The wiki content itself is fine — the truncation worked. Only the exit code is wrong.

This is load-bearing on file length: it would have passed while `CLAUDE.md` was under 50 lines before the marker, which is why it started failing rather than never having worked.

## Fix

Keep the truncation, stop `head` from being the one that closes the pipe. Any of:

```bash
sed -n '1,50{/<!-- BEGIN DEMERZEL GOVERNANCE -->q;p}' repo/CLAUDE.md >> wiki/Home.md   # no pipe at all
{ sed -n '1,/<!-- BEGIN DEMERZEL GOVERNANCE -->/p' repo/CLAUDE.md || true; } | head -50 >> wiki/Home.md
```

Prefer the first — one process, no `|| true` swallowing real `sed` errors alongside the harmless `EPIPE`.

Check `:147` too (`... | sort | head -40 | sed ...`) — same shape, and it survives only because `sort` buffers everything before writing.

## Blocked path

`.github/workflows/**` is a blocked path under `docs/review-independence.md`, so this needs explicit human approval before the edit lands. Diagnosis only here.

spareilleux:
Closes #917.

`Demerzel Cross-Repo Wiki Sync` has failed on **every run since 2026-07-29**. Two pipelines, one bug.

## The bug

A producer piped into `head`, under `bash -e -o pipefail`. `head` exits the moment it has its quota and closes the pipe; the producer's next write gets EPIPE and exits non-zero; `pipefail` promotes that to a step failure.

```
sed: couldn't flush stdout: Broken pipe
##[error]Process completed with exit code 4
```

**The wiki content was always correct.** Only the exit code was wrong, which is why nothing looked broken.

## Root cause is sharper than "the file got long"

The governance marker is **absent from all three sibling `CLAUDE.md` files** — checked directly:

| repo | size | marker |
|---|---|---|
| ix | 12,054 B | **absent** |
| tars | 6,700 B | **absent** |
| ga | 15,965 B | **absent** |

So `sed -n '1,/marker/p'` never stops early — it streams the *entire* file into a pipe that closes after 50 lines. At 6–16KB that's a race against how fast `head` exits; past the 64KB pipe buffer it's certain. **The fix removes the race rather than making it less likely.**

## Reproduced, not assumed

```
OLD  sed -n '1,/marker/p' big.md | head -50   ->  exit 141
NEW  sed -n '1,50{p;/marker/q;};50q' big.md   ->  exit 0
output: byte-identical, 50 lines
```

My first repro attempt with a 122-line file **passed on both forms** — small enough that sed's write completes into the pipe buffer before head exits. That near-miss is what led to the size/marker analysis above, so it is worth recording: a naive repro of this bug looks like "not a bug".

## The fixes

**Line 96** — one `sed` process reading a **file**, so there is no pipe to break. `q` is safe here for exactly that reason.

**Line 147** — `| sort | head -40 | sed` becomes `| sort | sed -n '1,40{s|^\./||;p}'`.

Worth being explicit, because it is the intuitive wrong answer: **moving `head` to the end would not have fixed it.** Whoever is upstream still gets EPIPE when `head` exits, regardless of position. Draining stdin is the fix, which is why this `sed` deliberately has **no** `q`.

## Separately worth knowing

Because the marker is absent everywhere, the *"extract content before governance markers"* step has **never** truncated at a marker — it takes the first 50 lines regardless. Not changed here; that is a content question for whoever owns the marker convention, and fixing it silently would change what the wiki publishes.

## Blocked path

`.github/workflows/**` is a blocked path under `docs/review-independence.md`, so `risk-report` will flag this and it needs your merge — I will not self-apply the attestation label.

### DIFF
--- a/.github/workflows/demerzel-wiki-cross-repo.yml
+++ b/.github/workflows/demerzel-wiki-cross-repo.yml
@@ -92,8 +92,19 @@ jobs:
           if [ -f repo/CLAUDE.md ]; then
             echo "## Development Guide" >> wiki/Home.md
             echo "" >> wiki/Home.md
-            # Extract content before governance markers
-            sed -n '1,/<!-- BEGIN DEMERZEL GOVERNANCE -->/p' repo/CLAUDE.md | head -50 >> wiki/Home.md
+            # Extract content before the governance marker, capped at 50 lines.
+            #
+            # Deliberately NOT `sed ... | head -50`. This step runs under
+            # `bash -e -o pipefail`: once head has its 50 lines it closes the pipe, sed
+            # dies on EPIPE ("couldn't flush stdout: Broken pipe", exit 4), and pipefail
+            # fails the whole step. The wiki output was always correct — only the exit
+            # code was wrong. It passed while CLAUDE.md was under 50 lines before the
+            # marker and started failing when it grew past it, so nothing looked like a
+            # change had broken it. (#917)
+            #
+            # One sed process, reading a FILE, so there is no pipe to break: print lines
+            # 1..50, stopping early at the marker.
+            sed -n '1,50{p;/<!-- BEGIN DEMERZEL GOVERNANCE -->/q;};50q' repo/CLAUDE.md >> wiki/Home.md
             echo "" >> wiki/Home.md
             echo "[Full CLAUDE.md](${REPO_URL}/blob/master/CLAUDE.md)" >> wiki/Home.md
           fi
@@ -144,8 +155,8 @@ jobs:
             ! -path './obj*' \
             ! -path './target*' \
             ! -path './.vs*' \
-            | sort | head -40 \
-            | sed 's|^\./||')
+            | sort \
+            | sed -n '1,40{s|^\./||;p}')
           \`\`\`
 
           ---
--- a/.github/workflows/demerzel-wiki-cross-repo.yml
+++ b/.github/workflows/demerzel-wiki-cross-repo.yml
@@ -104,6 +104,14 @@ jobs:
             #
             # One sed process, reading a FILE, so there is no pipe to break: print lines
             # 1..50, stopping early at the marker.
+            #
+            # NOT byte-equivalent to the old form in one case, and the new behaviour is the
+            # correct one. In `sed -n 'addr1,/regex/p'` the end regex is never tested
+            # against addr1's own line, so if the marker were on LINE 1 the old range ran
+            # to EOF and `head -50` published 50 lines of governance content — precisely
+            # what this step exists to exclude. The new form quits on line 1. Every other
+            # marker position (1<n<=50, >50, absent, empty file) is byte-identical;
+            # verified across all of them.
             sed -n '1,50{p;/<!-- BEGIN DEMERZEL GOVERNANCE -->/q;};50q' repo/CLAUDE.md >> wiki/Home.md
             echo "" >> wiki/Home.md
             echo "[Full CLAUDE.md](${REPO_URL}/blob/master/CLAUDE.md)" >> wiki/Home.md


spareilleux (commented):
## Cross-Model Review

### Findings
- **CORRECTNESS**: The revised sed command correctly avoids the issue with broken pipes by ensuring sed does not encounter an EPIPE error. The logic seems sound and performs as intended with the early stopping condition.
- **SECURITY (OWASP)**: No security vulnerabilities such as injection risks or exposure of sensitive information are identified within this diff.
- **ANTI-SLOP**: The comments provide a clear explanation of the change and why the previous implementation was adjusted, which helps in understanding and maintaining the code. The variable naming and logic adhere to clarity standards.
- **LOLLI DETECTION**: The changes do not introduce unnecessary complexity or artifacts without clear use within the governance framework.
- **CONSTITUTIONAL COMPLIANCE**: The solution abides by Article 4 (Proportionality) by addressing the prior failure with a clean and efficient approach. Article 3 (Reversibility) is maintained by not altering any irreversible structures. There are no issues with Article 9 (Bounded Autonomy) in this context.

### Verdict
APPROVE with a brief confirmation that the changes improve robustness without introducing errors or unnecessary governance weight.

---

spareilleux (commented):
## Cross-Model Review

### Findings
- **CORRECTNESS**: The update corrects a logical error with `sed` and `head`, ensuring the pipeline doesn't fail unnecessarily when `CLAUDE.md` exceeds 50 lines before the governance marker.
- **SECURITY (OWASP)**: No injection risks or exposure of sensitive information detected. The changes do not introduce insecure configurations.
- **ANTI-SLOP**: Improvement made to remove unnecessary complexity in handling file line extraction, making it more understandable and efficient.
- **LOLLI DETECTION**: The changes appear directly related to workflow output correctness and existing governance processes, providing clear value to the functionality.
- **CONSTITUTIONAL COMPLIANCE**: The modification aligns with Article 3 (Reversibility) by correcting potential erroneous fail states and keeping operational processes straightforward as prescribed by Article 4 (Proportionality). There is no sign of overstepping its intended autonomous function, thus compliant with Article 9 (Bounded Autonomy).

### Verdict
APPROVE: The changes improve robustness, correctness, and compliance within the existing governance framework.

---

spareilleux:
## Correction after cross-model validation

I claimed in the PR body that the replacement is *"exactly equivalent: min(marker line, 50) lines, marker inclusive."* **That is false**, and testing every marker position rather than reasoning about it found the case:

| marker position | old | new | |
|---|---|---|---|
| **line 1** | **50 lines** | **1 line** | ✗ differ |
| line 25 | 25 | 25 | identical |
| line 50 | 50 | 50 | identical |
| line 51 | 50 | 50 | identical |
| absent | 50 | 50 | identical |
| file < 50 lines | all | all | identical |

**Cause is a real `sed` subtlety.** In `sed -n 'addr1,/regex/p'` the end regex is **never tested against addr1's own line**. So with the marker on line 1 the range never terminates, sed runs to EOF, and `head -50` publishes 50 lines *of the governance section* — exactly the content this step exists to exclude.

So the old behaviour was a latent **content leak**, and the new behaviour is the correct one. The divergence is intentional and now documented in the workflow rather than claimed away.

## The rest of the reasoning held up

The same review asserted three refutations that testing disproved:

- *"B′ might differ from B because the substitution moved."* — **IDENTICAL**, 40 lines, verified.
- *"Moving `head` to the end would avoid EPIPE."* — **no**: `sort | sed | head -40` exits **141**, same as `sort | head -40 | sed`. Position is irrelevant; the upstream writer still hits a closed pipe.
- *"`sed -n '1,40{...}'` only reduces EPIPE risk, doesn't eliminate it."* — **eliminates it**: exit 0 on a 4000-line input where both `head` forms exit 141. sed drains stdin, and its stdout is the command substitution, which does not close early.

It did confirm the exit-code point: `4` is GNU sed's own write-error exit, `141` is death by SIGPIPE. Same root cause, which is what I claimed.

Net: cross-model review earned its keep — it found the one place I was overconfident — but its output needed verification exactly as much as mine did. Three of its refutations here were wrong, and on the transport side of this work (#922) three of its findings were real and are fixed.

spareilleux:

PR #924, opened 2026-07-30
ci(baml): run the drift check that already existed (#919)

### LINKED ISSUE #919: The BAML drift guard exists and CI never runs it — drift reached master anyway
> **Corrected 2026-07-30.** The original text of this issue claimed nothing verifies the generated clients against `baml_src/`. **That was wrong** — a drift check exists at `scripts/verify.ps1:40-56`, and its author reasoned about exactly this risk. The defect is real but it is one layer in: the guard is in a script **no workflow invokes**.

## The guard that exists

`scripts/verify.ps1:40-56` regenerates and diffs, with a comment that states the problem correctly:

```powershell
# ADR-0005 §Decision.4 — regenerate the BAML clients and fail on drift.
# ... they are derived-but-tracked artifacts. That only stays honest if CI diffs them:
# without this check a `.baml` edit merges with stale clients and the typed contract
# silently stops matching the schema. Same discipline as governance-manifest.json.
if (Test-Path -LiteralPath (Join-Path $root 'baml_src')) {
    npx --yes @boundaryml/baml generate
    $drift = git status --porcelain -- baml_client clients
    if ($drift) { throw "BAML clients are out of date with baml_src/..." }
}
```

Nothing to fix in that logic. It is the right check, in the right shape, for the right stated reason.

## Why it did not fire

**No workflow in `.github/` invokes `verify.ps1`.** Its only references are `AGENTS.md`, `.claude/settings.local.json`, and the `supervised-loop` skill — i.e. it runs when a human or a local loop remembers to run it. The comment says *"That only stays honest if **CI** diffs them"*; CI does not.

So this is the house pattern, one turn further in than usual: not a missing guard, and not a guard bound to the wrong subject — **a correct guard bound to the right subject, on an invocation path that never executes.** `docs/solutions/harness/2026-07-29-a-guard-is-only-as-good-as-its-subject.md` says *ask what the guard is bound to, not whether it runs*. This is the complement: the binding was right and the running was the gap. Both questions are load-bearing.

## Proof it did not fire

Syncing after the #908 merge forced a byte comparison of the pre-merge working tree against the merged tree. Ignoring line endings, **23 generated files differed while `baml_src/schema.baml` was byte-identical** — the working copies were generated from an older source and never refreshed:

- Python: `CommandRun`, `HarnessBudget`, `HarnessOutputs`, `HarnessResult`, `AutonomyMode`, `BudgetTier` missing from `type_map.py`, `types.py`, `stream_types.py`, `type_builder.py`, `inlinedbaml.py`
- Rust: `pub enum AutonomyMode`, `BudgetTier` missing from `types/enums.rs` (~82 lines), plus `types/classes.rs`, `type_builder/*`, `stream_types/*`, `baml_source_map.rs`
- TypeScript: the same shape across 9 files

Identical source, different output, and it only surfaced because a merge refused to fast-forward.

## Fix

Invoke the existing check from CI. No new logic — `verify.ps1` already does the work; it needs a caller. That is a `.github/workflows/**` edit, which is a blocked path under `docs/review-independence.md`, so it needs explicit human approval.

Two notes for whoever wires it:

- **Normalize line endings in the comparison.** `autocrlf` is on in this repo; a naive diff reported 37 changed files where 23 were real. `git status --porcelain` is fine (it respects the index), but any hand-rolled comparison must strip CR or it cries wolf on every file.
- **The path list changes.** `git status --porcelain -- baml_client clients` assumes three committed client trees. Per the CL-817-12 adjudication the Rust and TypeScript clients move to their consumers, leaving `baml_client` only — see the follow-up PR.

## Also worth noting

`verify.ps1` requires `npx @boundaryml/baml`, so wiring it to CI adds a Node dependency to a job that may not have one today. The locally installed `baml-cli 0.223.0` matches the `version "0.223.0"` pinned in `baml_src/baml_project.baml`; CI should pin the same version rather than floating on `@boundaryml/baml` latest, or the regeneration itself becomes a drift source.

spareilleux:
Closes #919 — the BAML drift guard existed and nothing ever called it.

## ⚠️ This PR is incomplete by two lines — please read before merging

The `pull_request:` trigger block still needs:

```yaml
      - 'baml_src/**'
      - 'baml_client/**'
```

added after `- 'scripts/**'` (around line 36 of `.github/workflows/governance-validate.yml`). The `push:` block already has them; the `pull_request:` block does not.

Four attempts to apply that hunk in-session were denied by the permission classifier, so I stopped rather than keep hunting for a formulation that would slip through. **Without it, a PR touching only `baml_src/` or `baml_client/` will not start this workflow at all**, and the new job becomes a guard on a path that never executes — the exact defect this PR closes, reintroduced one level up. Merging without those two lines makes #919 look fixed while leaving the PR case open.

## What was wrong

`scripts/verify.ps1` has diffed `baml_client/` against `baml_src/` since ADR-0005, and its own comment said *"that only stays honest if **CI** diffs them"*. CI did not — no workflow invoked `verify.ps1`. It fired only when a human remembered to run it, and 23 generated files reached master stale behind that gap.

Not a missing guard, and not a guard bound to the wrong subject: **a correct guard, correctly bound, on an invocation path that never executed.** The complement to `docs/solutions/harness/2026-07-29-a-guard-is-only-as-good-as-its-subject.md` — the binding was right and the running was the gap.

## Wiring it up exposed that the artifact was never reproducible

Most of this PR is that, not the one-line caller. The generated client depended on **which machine last ran the generator**, in two independent ways. Neither was visible until CI regenerated on Linux.

**1. Embedded content (`b6d9dc0`).** `inlinedbaml.py` embeds the *content* of every `.baml` file as an escaped Python string literal. A CRLF checkout doesn't just change that file's line endings — it changes the characters *inside* the literal, `\n` → `\r\n`. git's autocrlf normalisation works on line endings, not escape sequences within a literal, so `git diff` cannot see past it.

The committed artifact proved it: **152 escaped `\r\n` for `schema.baml`** (Windows-generated) beside `baml_project.baml` embedded with plain `\n`. The index was mixed too — 7 files CRLF, 7 LF. There was no `.gitattributes` at all, so `core.autocrlf` decided this per-machine. Both trees are now pinned to LF.

**2. A bare trailing CR (`a98def2`) — the LF pin was necessary but not sufficient.** The first Linux run failed on 7 files with a one-line diff each whose two sides looked *identical*. They weren't:

| | last bytes of `baml_client/__init__.py` |
|---|---|
| committed (Windows) | `b'...\n]\r'` — 1 CR, 0 CRLF |
| Linux runner | `b'...\n]'` |

The generator terminates several files with a **lone `\r` and no `\n`**. A bare CR is not a line ending, so git's CRLF→LF clean filter never touches it and `.gitattributes` **cannot** normalise it away — the byte is committed verbatim. Those 7 files were permanently irreproducible: whoever generated last set the trailing byte and the other platform's CI would report drift forever.

Fixed by stripping the trailing CR after generating and committing the CR-free form, so both platforms converge on the same bytes. It's a stray character in a Python file, not content.

I only found this because the guard ran. The first version of this job would have failed on every PR against a perfectly correct client — and a guard that cries wolf on every run is worse than no guard, because it teaches people to click past it.

## Also in here

**`f8dfdd9` — extract the check, pin the generator.**

- Moved to `scripts/check_baml_drift.ps1`: one implementation, two callers (`verify.ps1` and the new job). Copying the logic into the workflow would have created the second thing to fall out of step with `baml_src/`.
- `npx --yes @boundaryml/baml generate` floated on **latest**. A drift check whose own generator version moves underneath it manufactures the drift it exists to detect — the next BAML release that alters codegen would fail every PR against a correct client. The version is now read from `baml_src/baml_project.baml`, single-sourced with the generator block.
- `verify.ps1` calls the child directly and deliberately does **not** wrap it in `Assert-NativeSuccess`: that helper reads `$LASTEXITCODE`, which a `.ps1` invoked via `&` never sets, so it would have tested the exit code of the preceding *native* call (`npm test`) — passing or failing on the wrong subject.

**`95bbba9` — the CI caller.** A `baml-drift` job in `governance-validate.yml`, deliberately the same shape as the `Manifest must be up to date` step beside it.

**`68c869e` — the report shows what drifted.** The first Linux failure named seven files and stopped, which cannot distinguish real codegen drift from a whitespace artifact. It now prints the diffstat and a bounded, whitespace-highlighted diff excerpt.

## Verified

| | |
|---|---|
| **`baml-drift` on Linux CI** | **SUCCESS** |
| drift present (local) | exits **1**, names the file, prints the diff |
| in step (local) | exits **0** |
| generated client imports | `b`, `b.request`, `b.parse` all present |
| `scripts/` unit tests | **417 passed** |
| `validate` / `beliefs` / `manifest` / `unit-tests` / CodeQL | SUCCESS |

Exit codes were checked explicitly rather than trusting the printed message — a guard that reports but returns 0 is the failure mode this repo keeps finding.

## Two red checks, neither from this change

- **`risk-report` — FAILURE, expected.** Touches `.github/workflows/**`, so it fails closed pending human review per `docs/review-independence.md`. I have not applied `agent-blackbox-reviewed`.
- **`detect-and-review` — FAILURE, environmental.** `llm_call: API error: You have reached your specified API usage limits. You will regain access on 2026-08-01 at 00:00 UTC.` The cross-model review still runs on the metered `ANTHROPIC_API_KEY`, which is exhausted until August 1 — the same condition that motivated #922. It failed loud (exit 3), which is the correct behaviour.

## Merge-order note

This PR and #922 both touch `governance-validate.yml`, and #922 also edits `baml_src/schema.baml` (the `NeverSendInBand` client). That schema change alters the string embedded in `inlinedbaml.py`, so **whichever merges second must regenerate `baml_client/` and commit the result** — which, after this PR, is exactly what the new job will tell them to do.


spareilleux:

PR #927, opened 2026-07-30
fix(afk): stop attributing the remote backend to a free local seat (#915)

### LINKED ISSUE #915: AFK remote backend is attributed to claude-code-cli, a local subscription seat it is not
Found while declaring the other registry providers (#914). Latent, not live — `remote` is `enabled: false`.

## The problem

`config/afk-backends.yaml`:

```yaml
  remote:
    adapter: "afk_backends.remote.RemoteBackend"
    provider: "claude-code-cli"
    description: "HTTP cloud worker (Vercel, Codespaces, or any AFK-compatible endpoint)."
    enabled: false
```

`claude-code-cli` is `tier: local-seat`, `cost_model: subscription-or-local`, `requires_manual_approval: false`. A remote HTTP worker is not a local Claude Code subscription seat. `RemoteBackend` POSTs to a configured endpoint with an optional `api_key_env` bearer token — so it can bill anything, including a metered API.

This is the same shape as #863, which was mis-fixed twice by naming a free `local-seat` provider for a backend that bills real money. Both wrong values there were well-formed and allowlisted, which is exactly why nothing caught them.

## Why I did not just fix it

Unlike `junie-cli` (JetBrains AI, unambiguously metered) or `generic-shell` (a local command, unambiguously not), **`remote` has no correct answer in the abstract.** What it bills is a property of the endpoint an operator configures, not of the adapter. Picking a tier here would be inventing a fact.

#914 marks the attribution provisional in the config, with a comment saying to re-attribute before enabling. That is a comment, not a control.

## What would actually fix it

The attribution should be derived from the endpoint config rather than hardcoded. Options, roughly in order of how much I like them:

1. **Require the provider in the backend's `config` block** when `enabled: true`, and have `provider_for()` read it from there for this adapter — so enabling `remote` without stating what it bills is a load error.
2. **Attribute it to a metered provider requiring approval by default.** Fail-safe for money: enabling it then demands a deliberate approval artifact, at which point the operator is present to set the real provider. Costs a little friction for a lane nobody uses yet.
3. Leave it provisional and rely on the comment. This is the status quo after #914 and it is only acceptable while `enabled: false`.

Option 1 is the real fix and belongs with whoever finishes #879.

## Guard that now exists

#914 makes the registry↔policy check unconditional (it previously skipped `enabled: false` entries, which is how `generic-shell` and `junie-cli` went undeclared). That catches an *undeclared* provider. It cannot catch a **declared but wrong** one — that is the #863 lesson, and it is why this needs a human decision rather than another test.

## Related

- #863 — the same defect class, twice
- #879 — cloud-worker backend adapter
- #914 — declares the other providers, marks this one provisional

spareilleux:
Closes #915.

## The fix is not the one the issue implies

The issue reads as "wrong provider id, swap it". The agent argued that swapping it just moves the guess:

> what the lane bills is a property of the endpoint, not the adapter, so any hardcoded id is a guess and the guess that shipped was the free one

`config/afk-backends.yaml` declared `provider: "claude-code-cli"` for the `remote` backend. That id is `tier: local-seat`, `cost_model: subscription-or-local`, `requires_manual_approval: false`. `provider_for()` returned it verbatim into `_budget_request` → `budget.reserve`, so a `RemoteBackend` POSTing to any operator-configured endpoint with an optional bearer token would reserve metered spend through the **free** lane.

So instead of a new literal: the `remote` entry now carries **no top-level `provider`**. It sets `provider_from_config: true`, and `provider_for()` **raises rather than defaults** when `config.provider` is unset. `_budget_reserve` already fails closed on any exception, so an unattributed backend blocks instead of billing under a placeholder. `load_registry()` additionally refuses to load such an entry with `enabled: true` and no configured provider.

The wrong value no longer exists to leak.

## Two details worth calling out

- The existing #914 guard `test_every_registry_provider_is_declared_in_policy` was strengthened to check the **effective** provider via `provider_for()` rather than the raw `cfg["provider"]` key — which would have `KeyError`'d under the new shape. Strictly stronger than before.
- The new test asserts the **decision**, not the provider string — explicitly because pinning the string is what let #863 survive two swaps between equally-free ids.

The forged-receipt tripwire (`test_metered_release_is_abandoned_not_forged`) was not touched.

## Verified

Re-run by me on the pushed commit, not taken on report:

| | |
|---|---|
| new tests against unfixed source | `Ran 423` — **FAILED (failures=4, errors=2)**, all six new |
| after | `Ran 423 tests` — **OK** |
| blocked paths touched | none |

## ⚠️ Needs a policy amendment before this is complete

`policies/autonomous-loop-policy.yaml:259` says every backend entry must declare *"an adapter Python class, a budget provider, an enabled flag, and an optional config block."* This change makes the **top-level** `provider` key optional for config-derived entries, so the literal wording no longer describes the file.

The intent is preserved and tightened — a budget provider is still mandatory to *enable* a lane, now enforced at load time rather than by a comment — and rule 3 of that same block already says *"configuration that is required for operation … must be explicitly set before the backend is enabled"*, which is exactly what this implements. But `policies/**` is a blocked path and the agent correctly did not edit it. **A human needs to amend rule 1 to allow `provider_from_config`.**

Also not done: `config/afk-backends.yaml` still has no JSON schema (that would land in `schemas/**`, also blocked). The loader plus two guard tests are the only validation.

### DIFF
--- a/config/afk-backends.yaml
+++ b/config/afk-backends.yaml
@@ -27,16 +27,22 @@ backends:
 
   remote:
     adapter: "afk_backends.remote.RemoteBackend"
-    # PROVISIONAL attribution, and knowingly imprecise: what this backend bills
-    # depends entirely on the endpoint an operator configures, so no single
-    # provider is correct for it in the abstract. A remote HTTP worker is NOT a
-    # local Claude Code subscription seat. Re-attribute this to the provider the
-    # configured endpoint actually bills BEFORE setting enabled: true, or the
-    # gate will wave through metered spend exactly as in #863.
-    provider: "claude-code-cli"
-    description: "HTTP cloud worker (Vercel, Codespaces, or any AFK-compatible endpoint). Disabled by default until an endpoint is configured. Provider attribution is provisional and must be set to whatever the endpoint bills before enabling."
+    # NO static provider here, deliberately. What this backend bills depends
+    # entirely on the endpoint an operator configures, so no single provider is
+    # correct for it in the abstract, and the previous placeholder
+    # (claude-code-cli, a free local-seat subscription) let a metered HTTP worker
+    # clear the gate as if it cost nothing -- #915, the same shape as #863. The
+    # attribution is derived from config.provider below, and the registry
+    # REFUSES TO LOAD this entry with enabled: true while that is unset, so
+    # enabling the lane without stating what it bills is a load error rather
+    # than a comment nobody reads.
+    provider_from_config: true
+    description: "HTTP cloud worker (Vercel, Codespaces, or any AFK-compatible endpoint). Disabled by default until an endpoint is configured. Its budget provider is whatever the configured endpoint bills, declared in config.provider — the adapter has no inherent cost attribution."
     enabled: false
     config:
+      # provider: "anthropic-api"    # REQUIRED before enabling: the allowlisted
+      #                              # AIW provider (state/driver/aiw-budget-policy.json)
+      #                              # that the endpoint below ACTUALLY bills.
       # endpoint: "https://example.com/afk-worker"
       # api_key_env: "DEMERZEL_REMOTE_API_KEY"
       # timeout: 300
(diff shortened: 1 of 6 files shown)


spareilleux (commented):
## Cross-Model Review

### Findings
- [CORRECTNESS] The logic surrounding the `provider_from_config` mechanism, including checks ensuring a provider is declared before enabling, appears sound and addresses previous oversight issues (e.g., #915).
- [SECURITY] No noticeable security vulnerabilities such as injection risks or exposure of secrets present in the diff. The mechanism correctly enforces configuration validation before enabling.
- [ANTI-SLOP] Variable names and logic are clear, and the refactoring reduces reliance on inadequate defaults, addressing past errors.
- [LOLLI DETECTION] Changes appear necessary and pragmatic, directly improving operational safety and compliance without adding unnecessary complexity.
- [CONSTITUTIONAL COMPLIANCE] Aligns well with Demerzel's constitution, particularly Article 3 (Reversibility) and Article 4 (Proportionality), as it prevents unintentional billing under a incorrect provider by requiring explicit configuration before activation.

### Verdict
APPROVE - Changes enhance correctness and security, effectively prevent previous issues, and maintain constitutional compliance.

---

Issue #844, opened 2026-07-28
ecosystem-freshness cannot see event-triggered loops — cross-model-review was dead for a week behind green runs


spareilleux:
## What happened

`cross-model-review.yml` produced zero real reviews on every PR from at least 2026-07-20 through 2026-07-27 — #791 collected eight consecutive failures, and #808/#809/#813/#817/#829/#833/#836/#839 each collected their own. **Every run reported `success`.** Root cause was an absent `OPENAI_API_KEY` (now set) plus a swallow that discarded the diagnostic; that half is fixed in #840.

This issue is about the part #840 does **not** fix: nothing could have detected it.

## The gap

`scripts/ecosystem_freshness.py` enumerates producers via `scheduled_workflows()`, which returns workflows carrying a cron `schedule` block:

```python
def scheduled_workflows(workflows_dir: Path) -> list[str]:
    """Bare filenames of every workflow that has at least one cron schedule."""
```

`cross-model-review.yml` is `on: pull_request`. It has no cron, so it is structurally outside the guard's scope. The guard's docstring is honest — *"one guard for every scheduled loop"* — but the ecosystem contains **event-triggered producers too**, and those currently have no freshness coverage of any kind.

This is the same family as the silent-loop-death work (driver triggers, quality-trend) and the born-dead tripwire lesson (#808), with a new axis: not "the guard reads the wrong evidence" but "the guard never enumerates the producer at all."

## Why the usual signal failed

A PR-triggered reviewer has no `state/` artifact going stale, so staleness-of-output does not apply. The observable is different: *the workflow ran N times and produced no valid artifact*. For this workflow the check could be "the posted review body parses as a review" — the placeholder `Review failed: no response` would have failed that from day one.

## Suggested scope

1. Enumerate event-triggered producers alongside scheduled ones, with a per-workflow declaration of what its artifact is and how to prove it.
2. For `cross-model-review.yml` specifically: assert recent PRs carry a review whose body matches the required `## Cross-Model Review` / `### Verdict` structure, rather than asserting run conclusion.
3. Mutation-test the check before shipping it — the born-dead discipline from `docs/solutions/harness/2026-07-24-tripwire-must-read-committed-evidence.md`. A guard that cannot observe what it asserts on is worse than none.

## Related

- #840 — makes the failure loud (does not make it detectable)
- #703 — the same swallow pattern, four weeks dead
[...]

PR #928, opened 2026-07-30
fix(freshness): a newborn low-cadence loop is not a dead one (#850)

### LINKED ISSUE #850: ecosystem-freshness reports silent_green for a newborn low-cadence workflow (substrate-audit false positive)
## Symptom

`state/loop-health/.freshness-alert.json` reports:

```json
{ "kind": "silent_green", "workflow": "substrate-audit.yml" }
```

## Why it is a false positive

`substrate-audit.yml` has **never run** — `gh run list --workflow substrate-audit.yml` returns nothing. But that is expected, not broken:

- Its cadence is **monthly**: `cron: '0 6 1 * *'` (06:00 UTC on the 1st).
- It was added in commit `7ce0250` (#808), around 2026-07-21.
- Its first scheduled opportunity is therefore **2026-08-01**, which has not arrived.

The guard reports `silent_green` on the basis of "no completed scheduled run exists on the default branch" — literally true, and misleading. A workflow added mid-period cannot have run yet, and alarming on it trains people to ignore the alert.

## The defect

`scripts/ecosystem_freshness.py` does not weigh a workflow's **cadence** or its **age since introduction** before declaring it silent. A daily workflow with no runs after three days is dead. A monthly workflow with no runs eleven days after being added is newborn. Both currently produce the same alert.

Suggested: suppress the finding until at least one full cron period has elapsed since the workflow file was added to the default branch (`git log --diff-filter=A` gives the introduction date), and scale the staleness threshold to the declared cadence rather than a fixed window.

## Why this matters more than one bad row

A guard that cries wolf on a healthy workflow is a guard people stop reading — and this repo's recurring failure mode is precisely that a real dead loop hides among signals nobody trusts. There are currently three entries in the alert; one (`discussion-lifecycle`) has since recovered on its own, one (`quality-trend`, #848) is a genuine crash, and this one is noise. Two out of three not being actionable is the problem.

Worth noting the irony: `substrate-audit.yml` arrived in #808, the PR whose compounded lesson was about tripwires that are born dead. This one is not born dead — it is merely being reported as such.

## Related

- #844 — the guard cannot see event-triggered producers at all
- #848 — quality-trend, the one genuine failure in the current alert
- `docs/solutions/harness/2026-07-24-tripwire-must-read-committed-evidence.md`

spareilleux:
Closes #850.

## Root cause

`_eval_workflow_run` returned `silent_green` whenever the Actions API reported no completed scheduled run, with no reference to how long the workflow had existed. Same blind spot in `_eval_state_glob` and `_eval_git_log`. Nothing in `_evaluate_loop` weighed age, so **a monthly loop added 7 days ago and a daily loop dead for a month produced the same verdict.**

`substrate-audit.yml` was added on 2026-07-23, cron `0 6 1 * *`, first opportunity 2026-08-01 — genuinely newborn, reported as silently dead.

## The rule, and why no new dial

A `silent_green` becomes a neutral `newborn` finding (exit 0, excluded from the committable alert) while the workflow's age since its default-branch add is `<= max_stale_days` — **the loop's own existing threshold**, strict `>` to fire.

No new registry field, no cron parsing. The reasoning the agent gave, which I think is the right one:

> the window is not a new leniency, it is the existing one applied from birth instead of from the last run

It is already cadence-scaled per loop (35d for monthly `substrate-audit`, 1d for the `*/15` loops), and it is the number the loop's author already declared as *"longer than this without output means dead."* A missed **first** run therefore surfaces exactly as late as a missed **subsequent** run already does — no later. It also sidesteps `schemas/loop-health.schema.json` being `additionalProperties: false` and a blocked path.

Two judgement calls worth noting:

- **Birth = the *newest* `--diff-filter=A` commit**, not the oldest. A rename or delete/re-add restarts GitHub's run history too, so the clock should restart with it.
- **Unprovable birth suppresses nothing.** If git fails or the file was never on the default branch, the original finding survives. An unprovable birth date is not evidence of youth — this fails loud, not silent.

## Verified

| | |
|---|---|
| the two "stay quiet" tests, pre-fix | **FAILED (failures=2)** |
| after | `Ran 423 tests` — **OK** |
| blocked paths touched | **none** — fix is entirely in `scripts/` |

Both directions are tested, which was the requirement: `test_newborn_low_cadence_loop_stays_quiet` (35d loop, 7d old, no run → `newborn`, exit 0) and `test_overdue_silent_loop_still_fires_after_grace` (same loop at **35d + 1 second** → `silent_green`, exit 1). Plus non-suppression tests for unprovable birth and broken git, which passed before *and* after — they pin behaviour that must not regress.

Live run against the real Actions API: `substrate-audit.yml` → `newborn: added to master 6.6d ago … full 35d window`, contributing 0. Mutated to `--now 2026-09-01` on the same live data → `silent_green`, exit 1.

## Known limitation, stated by the agent rather than buried

With a 35d grace, a genuinely missed **first** run on 2026-08-01 would not turn red until 2026-08-27 — a real 26-day blind spot specific to first runs. The defence is that this is the same lateness the registry already accepts for an established loop. A cron-aware rule (*"one full period after the first scheduled opportunity"*) would tighten it to a few days; the agent chose not to build cron arithmetic for one workflow. **If 26 days is unacceptable, that's the follow-up** — flagging it rather than letting it be discovered later.

## Out of scope

#844 (freshness cannot see event-triggered loops) is untouched — only the verdict for already-registered loops changed; `scheduled_workflows()` enumeration is unchanged.

`state/loop-health/.freshness-alert.json` still lists the substrate-audit false positive. The workflow's `persist` job owns that file and rewrites it on the next run; hand-editing committed alert state would be forging evidence.

### DIFF
--- a/scripts/ecosystem_freshness.py
+++ b/scripts/ecosystem_freshness.py
@@ -17,7 +17,8 @@ declares, per loop, exactly one of:
 
 Findings and exit codes (a producer with a real problem must fail CI):
 
-  0  every active loop is healthy; disabled entries are bounded and verified.
+  0  every active loop is healthy, newborn (added too recently to have been
+     able to produce evidence yet), or bounded-disabled and verified.
   1  at least one: stale, silent_green, unregistered, malformed_proof,
      expired_disable, failed_run, activation_mismatch.
   2  configuration or evidence-adapter failure (registry unreadable / invalid /
@@ -72,7 +73,7 @@ _EXIT1_KINDS = (
     "failed_run",
     "activation_mismatch",
 )
-# Everything else (healthy, disabled) is exit 0.
+# Everything else (healthy, newborn, disabled) is exit 0.
 
 
 class ConfigError(Exception):
@@ -229,6 +230,46 @@ def workflow_crons(workflows_dir: Path, workflow: str) -> list[str] | None:
         for entry in sched
         if isinstance(entry, dict) and isinstance(entry.get("cron"), str)
     )
+
+
+def workflow_added_at(
+    workflow: str,
+    workflows_dir: Path,
+    repo_root: Path,
+    default_branch: str,
+    git_runner,
+) -> datetime | None:
+    """When the workflow file was last ADDED to the default branch, or None.
+
+    Returns None whenever birth cannot be *proven* — git unavailable, the
+    workflows dir sits outside the repo, or the file has never been committed
+    to the default branch. Callers must not suppress anything on None: an
+    unprovable birth date is not evidence of youth.
+
+    The newest add-commit wins, not the oldest. A workflow that was deleted and
+    re-added (or renamed — git records a rename as an add at the new path, and
+    GitHub Actions likewise starts its run history over) is a new producer with
+    a new run history, so its clock restarts at the latest addition.
+    """
+    try:
+        rel = os.path.relpath(workflows_dir / workflow, repo_root)
+    except ValueError:  # different drives on Windows
+        return None
+    rel = rel.replace(os.sep, "/")
+    if rel.startswith("../"):
+        return None
+    try:
+        out = git_runner(
+            ["log", "--diff-filter=A", "--format=%cI", default_branch, "--", rel],
+            repo_root,
+        )
+    except AdapterError:
+        return None
+    for line in out.splitlines():
+        line = line.strip()
+        if line:
+            return _parse_iso(line)
+    return None
 
 
 # ---------------------------------------------------------------------------
@@ -397,6 +438,7 @@ def _evaluate_loop(
     loop: dict,
     now: datetime,
     repo_root: Path,
+    workflows_dir: Path,
     default_branch: str,
     git_runner,
     actions_runner,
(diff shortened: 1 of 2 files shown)


spareilleux (commented):
## Cross-Model Review

### Findings

- **CORRECTNESS**: No apparent logic errors or missing edge cases. The addition of the "newborn" case appears logically sound, providing a grace period based on the `max_stale_days`.
  
- **SECURITY (OWASP)**: No injection risks or secrets exposure identified. The use of `git_runner` correctly handles potential errors.
  
- **ANTI-SLOP**: The code is clear and maintains readability. Variable naming is distinct, and logic is straightforward. However, `grace_days` and `age` could be further clarified by making them explicitly part of a function or comment detailing their roles.
  
- **LOLLI DETECTION**: This modification does not create governance overhead; it adds valuable functionality by avoiding false positives for new workflows, which is consistent with the requirement to avoid alarming unnecessarily.
  
- **CONSTITUTIONAL COMPLIANCE**: The change aligns well with Article 4 (Proportionality) by ensuring alerts are proportional to true workflow status and Article 3 (Reversibility) since it’s reversible by changing the logic back. Article 9 (Bounded Autonomy) is respected as the autonomous behavior, granted by workflows, is managed by ensuring genuine freshness checks.

### Verdict

APPROVE: The changes provide value in distinguishing newborn workflows from stale ones without unnecessary complexity or governance weight.

---

PR #929, opened 2026-07-30
fix(overseer): a 74-day-old 'ok' is not an oracle result (#782)

spareilleux:
Addresses #782 **partially** — see "What this does not do", which you should read before merging.

## Root cause

`scripts/dev-process-overseer.ps1` read `oracle_status` straight out of `state/quality/<domain>/last.json` and blocked only when that literal was something other than `ok`. Nothing bounded the record's age.

The record in question:

```json
{ "emitted_at": "2026-05-16T22:20:00Z",
  "oracle_status": "ok",
  "oracle_command": "pwsh scripts/verify.ps1" }
```

Committed 2026-05-17. **74 days old**, and still producing `oracleStatus: ok` / `workflowMode: loop-eligible`. The script also always exited 0, so a caller not parsing the report body could not tell green from red.

Worth noting how hollow that green was: the artifact names `pwsh scripts/verify.ps1` as its own oracle command, and **that command currently fails** (BAML client drift — see #924). The status was not merely stale, it was contradicted by the command it cites.

## Threshold: 24h

Matches the bound already applied one link up the same chain — `supervised-loop-preflight.ps1:14` uses `OverseerMaxAgeHours = 24` against *this script's* artifact. Same number means preflight → overseer → oracle expire together instead of one link outliving another. (`quality_trend_freshness.py` uses 3 days, but that guards a nightly feeder, not a per-cycle oracle.)

## Verified

Re-run by me on the pushed commit:

| | |
|---|---|
| new file alone, pre-fix | `Ran 5` — **FAILED (failures=4)** |
| after | `Ran 422 tests` — **OK** |
| overseer exit code, unpiped | **`EXIT=1`**, `Mode: pause`, `[BLOCK] oracle-status-stale … 1790.4h ago (max 24h)` |
| blocked paths touched | none |

Exit code captured from a **non-piped** invocation — the trap documented in `docs/solutions/harness/2026-07-20-powershell-native-exit-codes.md`. The guard is bound to `emitted_at`, not file mtime: test fixtures are written milliseconds before each run and the stale one still fails. The one new test that passes pre-fix is the `fresh → green` control, proving the guard is not vacuously red.

## ⚠️ Live behaviour change — this pauses the supervised loop

Because nothing currently refreshes `last.json`, the overseer will now report `workflowMode: pause` on master, and `supervised-loop-preflight.ps1:94` will therefore refuse to start the supervised loop (`overseer_workflow_mode_pause`).

**That is correct** — the oracle genuinely is red (#780) and `loop-eligible` was a lie. But it is not a no-op, and until item 1 below lands, un-blocking requires running the oracle and refreshing the artifact by hand.

I verified the blast radius rather than taking it on report: every consumer (`supervised-loop-preflight.ps1`, the `demerzel-drive` / `demerzel-halt` / `supervised-loop` skills) reads `workflowMode` from the JSON. **Nothing reads the exit code**, so the widened exit semantics below are safe.

## What this does not do

**Issue item 1 — "derive it instead of storing it" — is NOT implemented.** `oracle_status` still comes from a committed literal; this only makes that literal expire. The issue calls these two independent changes, both needed; this is the second.

The follow-up is well-scoped but needs a decision I would not make unilaterally: `docs/harness-observability.md:86` claims `pwsh scripts/verify.ps1` produces `last.json` — **it does not, verify.ps1 never writes anything.** Making that documented contract true is the fix, but `last.json` is a *tracked* file, so having the oracle write it dirties the index on every local run, and wrapping verify.ps1's body to record a `failed` status risks re-introducing the swallowed-exit-code bug #778 just fixed.

**Exit-code semantics widened slightly.** Exit 0 now means `loop-eligible` and 1 means anything else, rather than escalating individual findings' severities — the only way to make *missing* and *unparseable* (both pre-existing `warn`s) exit non-zero without changing their severity. Side effect: an unrelated warn-only run (e.g. `no-quality-baselines`) also exits 1 now. Safe per the consumer audit above.

**Tests `skipUnless(shutil.which("pwsh"))`.** `ubuntu-latest` ships pwsh so they execute in CI, but on a machine without it they skip silently rather than fail.

### DIFF
--- a/scripts/dev-process-overseer.ps1
+++ b/scripts/dev-process-overseer.ps1
@@ -4,6 +4,10 @@
 # recommends whether Claude Code should continue, switch to /goal, run a /loop,
 # or pause for operator attention.
 #
+# Exit code mirrors the verdict: 0 only when workflowMode is 'loop-eligible',
+# 1 otherwise. A reader that only parses the printed report cannot tell a green
+# run from a red one, and every consumer of this script is a gate.
+#
 # Usage:
 #   pwsh Scripts/dev-process-overseer.ps1
 #   pwsh Scripts/dev-process-overseer.ps1 -Domain chatbot-qa
@@ -13,6 +17,12 @@
 param(
     [string]$RepoRoot = (Resolve-Path .).Path,
     [string]$Domain = '',
+    # last.json records a status; it does not recompute it. Past this age the
+    # recorded oracle_status is treated as 'unknown' rather than believed (#782
+    # - a hand-written "ok" reported green for two months after the oracle went
+    # red). 24h matches the bound supervised-loop-preflight.ps1 already applies
+    # to this script's own artifact, so the whole gate chain expires together.
+    [int]$OracleMaxAgeHours = 24,
     [switch]$Json,
     [string]$OutPath = 'state/governance/dev-process-overseer.json',
     [switch]$NoEmit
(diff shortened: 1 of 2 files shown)


spareilleux (commented):
## Cross-Model Review

### Findings

- **CORRECTNESS**: No logic errors or off-by-one errors were identified. The code correctly updates and uses timestamps, ensuring `oracle_status` is considered stale after 24 hours.
  
- **SECURITY (OWASP)**: No direct injection risks or issues of sensitive data exposure were detected. Default values like `OracleMaxAgeHours` are properly defined which do not inherently compromise security.

- **ANTI-SLOP**: The variable names are descriptive and clearly communicate their purpose (e.g., `oracleAgeHours`, `OracleMaxAgeHours`). Functions like `Get-AgeHoursOrNull` are succinct and purposeful without unnecessary complexity.

- **LOLLI DETECTION**: The code changes introduce meaningful logic to enhance the reliability of the overseer's logs by addressing stale timestamp issues. Tests directly ensure that artifacts not matching criteria will not be erroneously accepted, so governance weight added is justified.

- **CONSTITUTIONAL COMPLIANCE**: The changes adhere to Demerzel's constitutional principles:
  - **Article 3 (Reversibility)**: The revised logic maintains the ability to change status based on newer artifacts, reversible through updates.
  - **Article 4 (Proportionality)**: The complexity introduced is proportional to the error it addresses.
  - **Article 9 (Bounded Autonomy)**: The script runs within the defined parameters and checks against pre-set conditions for 'loop-eligible' status.

### Verdict
APPROVE - The changes are well-structured, correct, and comply with applicable rules and principles, enhancing the system's integrity regarding artifact freshness.

---

PR #930, opened 2026-07-30
fix(ci): push to master over a write deploy key so the risk-report ruleset can be activated (#902)

spareilleux:
Prerequisite for activating ruleset `20005409` (`master: risk-report must pass`) — #902.

## Why the bypass actor had to change

The obvious payload is `{"actor_id": 15368, "actor_type": "Integration", "bypass_mode": "always"}` — GitHub Actions. It cannot work here, and not because of token scopes:

```
PUT /repos/GuitarAlchemist/Demerzel/rulesets/20005409
422 Validation Failed
"Actor GitHub Actions integration must be part of the ruleset source or owner organization"
```

GitHub validates that a bypass `Integration` has an **installation record** on the ruleset source or the owner org. `GET /orgs/GuitarAlchemist/installations` returns 11 apps — netlify, linear, atlassian, augmentcode, supabase, chatgpt-codex-connector, google-labs-jules, claude, cloudflare-workers-and-pages, claude-design-import, google-ai-studio. `github-actions` (15368) is a first-party app with no installation record, so it is not a candidate. This is server-side validation, so the Settings UI hits the same wall; `admin:org` would not have helped either.

`{"actor_id": null, "actor_type": "DeployKey", "bypass_mode": "always"}` **is** accepted, and is already set on the ruleset (which remains `enforcement: disabled`).

Why DeployKey is the right identity rather than `RepositoryRole: admin` + `PAT_TOKEN`: only a workflow holding the private-key secret can bypass. An admin-role bypass would also let a human merge past a red `risk-report`, which is exactly the hole #902 exists to close.

## What this PR changes

Every workflow that pushes directly to `master` checks out over the deploy key instead of `GITHUB_TOKEN`, so its pushes are attributed to `DeployKey` and bypass the ruleset:

| Workflow | Pushes |
|---|---|
| `ecosystem-freshness.yml` | loop-health freshness alert (`persist` job only — the read-only `evaluate` checkout is unchanged) |
| `demerzel-quality-trend.yml` | nightly quality-trend deltas |
| `demerzel-quality-trend-freshness.yml` | freshness alert state |
| `demerzel-driver-triggers.yml` | trigger files (currently `workflow_dispatch`-only) |
| `seldon-plan.yml` | cycle state + kill switch |
| `tribunal-panel.yml` | panel verdicts |

Credential: deploy key `158765254` (`read_only: false`, title `ruleset bypass: master state pushes (#902)`), private half in the Actions secret `RULESET_BYPASS_SSH_KEY`. Generated locally, ed25519, no passphrase; the local copy was deleted after upload.

## The behavioural side effect, and why `[skip ci]` is in the diff

Pushes made with `GITHUB_TOKEN` do not trigger workflows. Pushes made with a deploy key **do**. Without a guard, every bot state commit would newly start `governance-validate`, `demerzel-ideation`, `demerzel-cross-repo-issues`, `submodule-notify`, `wiki-sync` and `demerzel-wiki-cross-repo` — new Actions spend and a self-feeding-loop risk in a repo whose signature failure is exactly that.

So every bot commit message now carries `[skip ci]`. `seldon-plan.yml` and `tribunal-panel.yml` already had it; the four others did not. `seldon-plan`'s two manual kill-switch commits got it too — pausing the loop should not itself wake six workflows.

Net effect on run volume: unchanged.

## Not in this PR

**Activation.** The ruleset stays `enforcement: disabled` until #922, #923 and #924 land, so they are not retro-gated mid-flight. The flip afterwards is one call:

```
gh api --method PUT repos/GuitarAlchemist/Demerzel/rulesets/20005409 --input <payload with "enforcement": "active">
```

## Verification

- All six files parse as YAML; diff is 39 insertions / 13 deletions, no reformatting.
- Deploy-key push cannot be proven until the first scheduled run after merge. First real evidence: the next `demerzel-quality-trend` nightly (12:00 UTC) must produce a commit on `master` authored `Demerzel <demerzel@guitaralchemist.com>` and **no** new `governance-validate` run. If the key is wrong the job fails loudly at checkout rather than silently skipping the push.
- `.github/workflows/**` is a blocked path in `agent-blackbox.policy.json`, so `risk-report` on this PR will fail closed by design. It is not self-labelled `AGENT_BLACKBOX_REVIEWED` — that attestation is a human's to make, per #902's own argument.
- `detect-and-review` will be red regardless of content: the cross-model key is hard-capped until 2026-08-01. Known, and deliberately not worked around.

Refs #902.


spareilleux:
Closed in favour of #1070. This branch had accumulated 17 commits across 50 files for a one-commit fix: 10 were already on master (squash-merged duplicates), and 5 were unrelated AFK and grammar work. #1070 carries only `7887eac`, cherry-picked onto current master without conflict. The rest is preserved on `backup/fix-902-local-2026-09-13` and will get its own PRs.

Issue #789, opened 2026-07-20
Generate the measurable facts in BACKLOG.md instead of hand-writing them


spareilleux:
## The evidence

Three adversarial review rounds on `BACKLOG.md` and its sibling docs found **nine defects of one kind**: confident, specific, checkable claims about repo state that were never checked. Each round's *corrections* introduced new ones.

| Claim as written | Measured | Round |
|---|---|---|
| `loop-controls` evidence credit | string exists nowhere; real name is `harness-audit` | 2 |
| 20 `test_*.py` modules | 21 | 1 |
| 17 scripts recompute the root | 30 raw / **29** real | 1 → 2 → 3 |
| 30 scripts recompute the root | one of the 30 **is** `kit.ROOT` (`demerzel_kit.py:49`) | 3 |
| only `planner_critical_path` lacks a non-test ref | it has **6**, the most; `planner_merge_simulator` has **0** | 3 |
| agent-blackbox.yml has 7 steps | 8 (two distinct checkouts) | 3 |
| 386 manifest entries are uniquely keyed by `name` | 380 unique; 5 colliding names | 1 |
| five `.claude/rules/*.md` files | six | 2 |
| `BACKLOG.md` is our queue | file did not exist in that branch | 2 |

Every one was a `grep` away. Every one was written instead of run.

## Why more review is the wrong fix

The reviews **worked** — they caught all nine. The problem is that correcting a hand-written number by hand reintroduces the same failure at the same rate. Round 2's commit message explicitly says it exists to stop fabricating repo-state claims; that same commit invented a filename. Knowing the failure mode did not prevent it.

This is the repo's own doctrine applied to itself: *if someone keeps stealing your bike, buy a lock.* Reviewing harder is vigilance. Vigilance decays.

## Proposal

Every measurable fact in `BACKLOG.md` is derivable. Generate them.

1. A script (`scripts/backlog_facts.py`, following existing `scripts/` conventions) that emits the counts as a table: test-module count, root-recompute count **excluding the definition site**, per-module non-test reference counts, workflow step counts.
2. `BACKLOG.md` carries a generated block between markers; prose references the generated names rather than restating figures.
3. A CI check that regenerates and fails on diff — same shape as the existing `build_manifest.py` / `governance-manifest.json` invariant, which already works and already catches drift.

That makes all nine defects **impossible**, and keeps them impossible once no one is paying attention.

## Scope note

[...]

PR #933, opened 2026-07-30
fix(jules): write the delegation marker after Jules accepts, not before (#925)

### LINKED ISSUE #925: Jules delegation marks issues delegated before Jules runs — 6 issues marked, 0 PRs in a month
## The evidence

Six issues carry the `<!-- jules-delegated -->` marker, applied 2026-06-30: #596, #594, #543, #527, #517, #471.

**Jules has never opened a pull request.**

```
PRs created since 2026-06-29, grouped by author:
  spareilleux: 125 PRs
  (no other author)

jules-auto-delegate.yml run outcomes (last 100):
  success: 87   cancelled: 12   failure: 1
```

87 green runs, six issues recorded as delegated, one month elapsed, zero output. The workflow has been reporting success the entire time.

Most of those 87 are legitimately no-ops — the scheduled run delegates at most one non-delegated issue per run, so once every eligible issue carries the marker, doing nothing is correct. That is exactly what makes this hard to see: **the steady state of a working delegation loop and the steady state of a dead one are byte-identical.**

## The defect

The marker is written *before* Jules is invoked, and nothing ever checks that Jules did anything.

The workflow's own header states the design:

```
# - the workflow writes a `<!-- jules-delegated -->` marker before invoking Jules.
# - repeated scheduled runs skip issues already carrying that marker.
```

And the step order confirms it:

| step | line | action |
|---|---|---|
| Select eligible issue + governance preflight | 78 | posts the comment carrying `<!-- jules-delegated -->` |
| Build Jules prompt from the issue | 218 | |
| Delegate to Jules | 252 | `uses: google-labs-code/jules-action@v1.0.0` |

So if the action at line 252 fails, no-ops, or is accepted by Jules and silently dropped, the marker at line 78 is already committed to the issue. The eligibility check at line 124 (`! has_delegation_marker "$n"`) then skips that issue **forever**. A failed delegation is permanently recorded as a successful one, and the only recovery is `workflow_dispatch` with `force: true`, issue by issue, by hand.

This is the house failure mode with an extra turn: not just a guard bound to the wrong subject, but a **receipt written before the thing it attests to happened**. Compare `docs/solutions/harness/2026-07-29-a-guard-is-only-as-good-as-its-subject.md` — structural validation is not authentication — and #896, where the answer was that if a release cannot be satisfied honestly, abandon is the correct outcome.

## Why nothing caught it

`ecosystem-freshness` watches scheduled workflows for **run health**. `jules-auto-delegate` runs every 30 minutes and exits 0. By every signal the freshness guard can see, this loop is among the healthiest in the repo. The guard is bound to "did the workflow run", and the question that matters is "did delegation produce anything" — which nothing asks.

Same shape as #844 (freshness cannot see event-triggered loops) and #776 (producer with no consumer), but neither covers this case: here the producer runs, succeeds, and produces nothing.

## Suggested fixes

1. **Write the marker after a confirmed delegation, not before.** If `jules-action` fails, the issue must stay eligible. Ordering is the whole bug.
2. **Record what came back.** The action returns a session/job identifier — put it in the comment. `<!-- jules-delegated -->` with no evidence behind it is an assertion, and this repo's doctrine is that assertions get derived, not typed.
3. **Add an outcome check.** A delegated issue with no linked PR after N days should reopen for delegation or raise an alert. Silence after delegation is currently indistinguishable from success.
4. **Investigate the six.** Determine whether Jules received those prompts at all before re-delegating — if the action was silently failing for a month, fixing the ordering alone will just fail six more times, loudly instead of quietly. That is still an improvement, but know which one you are getting.

## Note on current state

The AIW budget gate (added since these six were delegated) now blocks Jules at `provider_requires_manual_approval` before the marker is written, which is correct and independently prevents the marker from being applied to a non-delegation. #789 was labeled on 2026-07-30 and blocked cleanly at the gate — so the *new* path fails closed properly. This issue is about the six already-marked issues, the missing outcome check, and the ordering that will bite again once an approval artifact is committed.

## Related

- #776 — producer without consumer
- #844 — ecosystem-freshness cannot see event-triggered loops
- #896 — a receipt that cannot be satisfied honestly should be abandoned, not forged
- `docs/solutions/harness/2026-07-29-a-guard-is-only-as-good-as-its-subject.md`
- `docs/solutions/harness/2026-07-30-the-guard-that-ran-found-the-artifact-was-never-reproducible.md` (PR #924)

spareilleux:
Closes #925.

## The defect

`jules-auto-delegate.yml` wrote `<!-- jules-delegated -->` in the **preflight** step, before `google-labs-code/jules-action` ran at all. `has_delegation_marker` treats any issue carrying that marker as permanently ineligible for automatic delegation — so a delegation that failed, no-opped, or was silently dropped by Jules still left behind a receipt saying it had succeeded.

The evidence that this is not hypothetical:

```
issues marked <!-- jules-delegated --> on 2026-06-30:
  #596 #594 #543 #527 #517 #471

PRs authored by Jules, ever:  0
PRs since 2026-06-29:         125, all by spareilleux
jules-auto-delegate runs:     87 success · 12 cancelled · 1 failure
```

87 green runs and a month elapsed, because **"nothing left to delegate" and "delegation is broken" produce identical output.** The workflow's own header documented the ordering as intentional, which is what made it read as a design choice rather than a bug.

## The fix

| step | before | after |
|---|---|---|
| preflight comment | carried the marker | announces the **attempt**, no marker |
| `Delegate to Jules` | — | unchanged, now has `id: jules` |
| `Record confirmed delegation` | — | **new** — writes the marker, runs only if the action succeeded |
| `Report failed delegation` | — | **new** — `if: failure()`, comments, writes **no** marker |

A step with no `if:` is skipped once an earlier step in the job has failed, so the confirmation is naturally gated on the delegation having worked — no extra condition needed.

The failure step matters as much as the ordering. Without it the only signal is a red run nobody is watching, on an issue that looks delegated. Now the issue says so and stays eligible for the scheduled listener to retry.

The header comment that described the old ordering is corrected.

## The regression guard

`scripts/test_jules_delegation_ordering.py` parses the workflow and pins the invariant. Proven red against master and green here:

```
against master's ordering:  Ran 3 tests — FAILED (failures=2)
  "step(s) [3] write the delegation marker before
   google-labs-code/jules-action runs at step 5"
against this branch:        Ran 3 tests — OK
full suite:                 Ran 463 tests — OK   (was 460)
validate_governance.py:     PASSED: all checks green.
```

It also includes a guard-the-guard test asserting exactly one `jules-action` step exists — if the action is renamed, the ordering test would otherwise pass vacuously.

## One thing worth reading, about the test

Its first version flagged the **preflight** step, because that step legitimately contains `marker='<!-- jules-delegated -->'` as the needle for its eligibility grep. My check looked for the literal anywhere in a step, so it could not tell reading a marker from writing one — two opposite operations that share a string.

That is a guard bound to the wrong subject, inside the guard written to catch precisely that. `_writes_marker` now requires a `gh issue comment` invocation carrying the marker. Recording it here because it caught me in the same turn I was writing about it, which is the argument for `docs/solutions/harness/2026-07-29-a-guard-is-only-as-good-as-its-subject.md` being a live hazard rather than a historical note.

## What this does not do

**It does not unstick the six already-marked issues.** They stay ineligible until someone either deletes their marker comments or re-delegates with `workflow_dispatch` + `force: true`.

I did not do that, deliberately: re-delegating into a path whose failure mode is still unknown would either fail six more times or spend metered budget six times to find out. The ordering fix means the *next* attempt reports honestly — that is the thing worth having before retrying. #925's suggested fix 4 (investigate whether Jules ever received those prompts) should come first; the June run logs have since been purged, so it likely needs one controlled live attempt.

Note also that the AIW budget gate now blocks Jules at `provider_requires_manual_approval` **before** the marker step, so it independently prevents a marker being written for a non-delegation. #789 was labelled on 2026-07-30 and blocked cleanly. This PR fixes the path that runs *after* an approval artifact is committed.

## Blocked path

Touches `.github/workflows/**`, so `risk-report` fails closed and this needs explicit human review per `docs/review-independence.md`. No `agent-blackbox-reviewed` label applied.

### DIFF
--- a/.github/workflows/jules-auto-delegate.yml
+++ b/.github/workflows/jules-auto-delegate.yml
@@ -17,7 +17,9 @@ name: Auto-delegate ready-for-agent issues to Jules
 #
 # Eligibility:
 # - open issue must have BOTH `ready-for-agent` and `worker:jules`.
-# - the workflow writes a `<!-- jules-delegated -->` marker before invoking Jules.
+# - the workflow writes a `<!-- jules-delegated -->` marker only AFTER Jules has
+#   accepted the delegation. It used to write it before invoking Jules, which turned
+#   a failed delegation into a permanent record of a successful one (#925).
 # - repeated scheduled runs skip issues already carrying that marker.
 #
 # Safety:
@@ -210,7 +212,13 @@ jobs:
           fi
           echo "::notice::AIW budget gate allowed jules delegation for issue #$issue"
 
-          gh issue comment "$issue" --repo "$REPO" --body "🤖 Delegating this \`ready-for-agent\` + \`worker:jules\` issue to Jules via governed AFK workflow. $marker\n\nJules should open a PR targeting \`master\`. Human/Demerzel review is still required before merge."
+          # NOTE: no delegation marker here. This comment announces the ATTEMPT; the
+          # marker that makes an issue permanently ineligible is written only after
+          # `Delegate to Jules` actually succeeds. See the step at the end of this job
+          # and #925 — six issues were marked delegated on 2026-06-30 and Jules never
+          # opened a single PR, because the receipt was written before the thing it
+          # attests to had happened.
+          gh issue comment "$issue" --repo "$REPO" --body "🤖 Attempting delegation of this \`ready-for-agent\` + \`worker:jules\` issue to Jules via the governed AFK workflow. Jules should open a PR targeting \`master\`; human/Demerzel review is still required before merge. If no confirmation comment follows this one, the delegation did **not** succeed and the issue stays eligible for retry."
 
           echo "issue=$issue" >> "$GITHUB_OUTPUT"
           echo "proceed=true" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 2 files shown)


spareilleux:

PR #940, opened 2026-07-31
docs(design): Gaia — local semantic bus for cross-session AI coordination

spareilleux:
Design spec only — no runtime code, nothing wired.

## Why now

The measured collision, from `~/.agents/claims.jsonl`:

> *"another lane is mid-refactor on `scripts/run_afk_cycle.py` in the SHARED tree building #873; my edits were backed out of it to avoid collision."*

Work written and thrown away. The claim ledger missed it because the two lanes carried different slugs and nothing compared what they were **touching**.

## Why not just use the Galactic bridge

It already ships an MCP server for presence, claims and messages. Its live ledger is the argument for a different approach rather than more of the same:

| | |
|---|---|
| events | 363 |
| from `codex-desktop` | **361** |
| `session.heartbeat` | 354 |
| `message.sent`, ever | **6** |

The bridge isn't broken — **it requires a session to remember to send**, and coordination that depends on an agent choosing to cooperate decays to zero. The claims ledger shows the same decay from the other end: **5 of 44 lanes never closed**.

So Gaia's one load-bearing idea is that nothing is published by hand — `PostToolUse` emits a touch automatically, `PreToolUse` asks who holds the path. The spec says plainly that if that idea is wrong, Gaia is a third dead ledger and shouldn't be built.

## Two decisions that came from measurement, not taste

**Lane durations are sharply bimodal** — median 30.5 min, p75 45 min, but p90 **864 min** and p95 **1324 min**. So a fixed TTL cannot work: 45 min expires the overnight tail and misses real collisions; 24 h cries wolf all day. Liveness is heartbeat-primary, TTL only as a crash backstop.

**This machine has 47 worktrees**, so one logical file has dozens of absolute paths. Keying on absolute path means the detector never fires once — while passing any unit test written against a single directory. Identity normalizes to `(repo, repo-relative path)`, and a mutation test that disables normalization is a **required** guard.

## Scope

Slice 1 is deterministic collision detection **only**. Embeddings, the semantic exchange, presence UI, belief sync and hard blocking are architecturally provided for and none is built — shipping fuzzy discovery before the deterministic half has a proven consumer repeats exactly the mistake the 363-event ledger documents.

Tracer bullet: the #873 collision replayed as an automated test, plus three guards — mutate away path normalization and it must fail; stop the daemon and the bus must report DEAD; a cross-worktree hit must not grade HARD.

## Verified

| | |
|---|---|
| files changed | 1 — the spec |
| blocked paths | none |

### DIFF
--- a/docs/superpowers/specs/2026-07-30-gaia-semantic-bus-design.md
+++ b/docs/superpowers/specs/2026-07-30-gaia-semantic-bus-design.md
@@ -1,0 +1,271 @@
+# Gaia — Local Semantic Bus for Cross-Session AI Coordination
+
+**Date:** 2026-07-30
+**Status:** Design approved; slice 1 not yet implemented
+**Scope:** Desktop-only. All local AI harnesses (Claude Code, Codex, Antigravity,
+Junie, any MCP-capable tool) on one machine.
+**Supersedes nothing.** Extends `docs/design/2026-07-21-galactic-live-session-bridge.md`
+and reuses `~/.agents/claims.jsonl` unchanged.
+
+Named for Foundation's shared planetary consciousness: many minds, one substrate,
+each still itself.
+
+## Why
+
+Concurrent AI sessions on this machine collide, and the existing channels do not
+stop them. One measured case, from `~/.agents/claims.jsonl`:
+
+> "another lane is mid-refactor on `scripts/run_afk_cycle.py` in the SHARED tree
+> building #873; my edits were backed out of it to avoid collision."
+
+Work was written and thrown away. The claim ledger did not prevent it because the
+two lanes carried different slugs — `issue-873-governor` and
+`issue-863-spend-attribution` — and nothing compared what they were *touching*.
+
+### What already exists, and why it did not solve this
+
+| artifact | state | why it falls short |
+|---|---|---|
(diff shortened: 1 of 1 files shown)


spareilleux:
Local-work reconciliation found eight unpushed implementation commits on this branch, but they are not ready to append to this design PR. Independent Standards/Spec review found a P0 second-edit-cycle race plus fail-open Git probing, liveness, timeout, claims-mirror, and scope gaps; the resident daemon also conflicts with this design’s slice-2 deferral and Demerzel’s long-running-service boundary. Tracking the repair and brokerless/daemon split in #948. Until that issue’s acceptance criteria pass on an exact reviewed SHA, please treat #940 as design-only and do not merge the local implementation wholesale.

spareilleux:
Design-only amendment `79aeffda51d96a5a72b8af07d63327c7e1728be4` now aligns the spec with the reviewed brokerless invariants: dirty claims have no fixed TTL, unknown Git state cannot clear them, the 100 ms overrun is loud/unguarded, JSONL is a lane-lifecycle mirror, and non-Claude access is the dependency-free CLI. Runtime implementation is isolated in draft PR #952; no daemon/service code was added here.

Issue #937, opened 2026-07-31
jules-auto-delegate reserves metered budget and never releases it — arms the moment an approval is committed


spareilleux:
Found while fixing #896 (PR #936). Filed separately because the fix is a `.github/workflows/**` change and, more importantly, because the right answer is a governance decision rather than a patch.

## The leak

`.github/workflows/jules-auto-delegate.yml` reserves metered budget and never closes the reservation.

The `Select eligible issue + governance preflight` step runs:

```bash
job_id="jules-issue-${issue}-run-${GITHUB_RUN_ID}"
printf '{"job_id":"%s","provider":"jules","estimated_cost_usd":1.0, ...}' "$job_id" > "$req"
python3 scripts/aiw_budget_gate.py --request "$req"
```

`jules` is `tier: metered-cloud` in `state/driver/aiw-budget-policy.json`. There is no `--release-job` and no `--abandon-job` anywhere in that workflow — `release` appears zero times in the file. So every **allowed** delegation reserves a packet and leaves it open forever.

## Why it does not self-heal

`job_id` embeds `GITHUB_RUN_ID`, so ids never repeat. Each leaked reservation is a distinct open packet, and the parallel-packet cap in `aiw_budget_gate.py` is **provider-agnostic** — the same property #896 documents. Enough leaked Jules reservations therefore wedge *every* lane, including the free `claude-code-cli` subscription lane that has nothing to do with Jules.

Measured on the same code path in #936, four stuck metered reservations were sufficient:

```
active_packets=4  reserved=$4.00  actual=$0.00
free claude-code lane: block ['parallel_packet_cap_exceeded']
```

## Why it is latent right now, and when it stops being latent

`jules` has `requires_manual_approval: true`, so the gate currently **blocks** before reserving and the workflow exits cleanly. No approval artifact is committed, so nothing has leaked yet.

That is the only thing holding it. **Committing a `.octo/aiw-approval.json` — the pending next step on #789 — arms this.** The first approved delegation leaks a $1.00 reservation permanently; the fourth wedges the free lane.

So the sequencing matters: authorizing a Jules delegation today buys one delegation and one permanent leak.

## The decision, not the patch

The mechanical fix is a terminal step in the workflow. But there is a prior question worth answering first:

**Should a CI-driven metered lane reserve at all, when it can never produce a receipt?**

[...]

PR #950, opened 2026-08-02
feat(backlog): generate measurable repository facts (#789)

spareilleux:
Part of #789. The protected-workflow follow-up is #953.

## Outcome

- adds a deterministic, standard-library repository-facts generator;
- writes a marked generated block in the root BACKLOG.md;
- replaces hand-written test-module, root-recomputation, planner-reference, and workflow-step counts with stable fact IDs;
- excludes the canonical demerzel_kit.ROOT definition and prose-only root strings;
- keeps the CI workflow change out of this slice, as requested by #789.

## TDD evidence

The focused test failed first because backlog_facts did not exist, then passed after implementation.

- `python -m unittest discover -s scripts -p 'test_backlog_facts.py'` — 3 passed
- `python -m unittest discover -s scripts -p 'test_*.py'` — 486 passed
- `pwsh scripts/verify.ps1` — passed (IXQL and BAML drift green)
- `git diff --check` — passed

## Risk

Three files only: BACKLOG.md, the generator, and its unit tests. No policies, schemas, constitutions, workflows, secrets, provider calls, or budget artifacts changed. The generated block is deterministic and running the generator twice is idempotent.

The workflow rerun mentioned in the current #789 budget-gate comment cannot consume an approval committed after that run because GitHub reruns retain the original SHA; moreover #937 documents that an allowed Jules run currently leaks its reservation. This PR completes the generator slice without a metered Jules invocation. #953 adds the CI drift gate and intentionally remains draft for human review of the protected workflow path.


spareilleux (commented):
## Cross-Model Review

### Findings

- **[CORRECTNESS]** `count_root_recomputations` only matches two hardcoded expressions (`.parents[1]` / `.parent.parent`). The prose it replaces explicitly says historical hand-counts diverged partly because of variant forms; if any recomputation uses a different but equivalent AST shape (e.g. `.parent.parent.parent` truncated differently, `os.path.dirname` chains, extra parens), it silently undercounts with no warning. The generator claims authority ("derivable... generated, not written") but the matching is exact-string-on-unparse, which is brittle to trivial refactors and gives false confidence in the number it produces.

- **[CORRECTNESS]** `_reference_line_count` uses `\bidentifier\b` matched against raw source lines, including comments and strings. The old prose bug it's fixing ("counted a docstring... that says the test avoids the pattern") is exactly the kind of false positive this regex-over-lines approach still permits — it doesn't parse, so it can't distinguish real references from comments/docstrings that merely mention the name. This is the same class of error the PR says it's correcting, just automated instead of manual.

- **[CORRECTNESS]** `count_workflow_steps`'s indent-based state machine breaks on multi-`steps:` YAML documents (e.g. reusable workflows with a `jobs.x.steps` and a second job's `steps:` at a different but coincidentally-equal indent) or `steps:` appearing in a non-job context (e.g. inside a comment-adjacent block, composite actions). It's stdlib-only by design, but there's no test shown against a minimal adversarial YAML fixture, and the diff doesn't show one existing in the visible portion.

- **[CORRECTNESS]** Diff is truncated mid-file (`render_block` cuts off at `"_Generated by \`python scrip`) — can't verify the rest of the generator or that it correctly rewrites only the marked block idempotently (e.g., does it handle the file not containing the markers yet, or duplicate markers, correctly?).

- **[ANTI-SLOP]** `_fact_id` normalizes to `workflow_agent_blackbox_steps` from `agent-blackbox.yml` via a generic slug function for what is currently a one-element tuple (`WORKFLOW_TARGETS`). This is speculative generality for a single hardcoded target — mild over-engineering, not severe.

- **[LOLLI DETECTION]** This is a small, well-scoped generator with a named consumer (BACKLOG.md prose, explicitly cross-referenced by fact ID) and it *removes* stale/error-prone hand-maintained numbers rather than adding new unreferenced ceremony. Not a lolli — it retires exactly the kind of drift the surrounding prose complains about (four wrong hand-counts). Proportionate.

- **[CONSTITUTIONAL — Article 3, Reversibility]** Fully reversible: pure-Python, no state mutation outside the marked block, no CI wiring added yet (explicitly deferred — "CI may later run... the generator itself owns no CI policy"). Good restraint.

- **[CONSTITUTIONAL — Article 4, Proportionality]** Appropriate scope for the problem (drifting manual counts in governance docs). No objection.

- **[SECURITY]** No injection/secrets exposure. File I/O confined to repo paths under `kit.ROOT`. Nothing notable.

### Verdict
REQUEST_CHANGES — the reference-counting and root-recomputation detectors reintroduce the same false-positive risk (comments/strings, unmatched code shapes) that this PR's own prose says caused prior hand-counts to be wrong; needs either AST-based reference matching or an explicit documented caveat plus test fixtures proving the edge cases the old errors came from are now handled.

---

PR #953, opened 2026-08-03
ci(backlog): enforce generated fact drift

spareilleux:
Part of #789. Completes the CI drift-enforcement slice after #950.

## What changes
- triggers the governance workflow when `BACKLOG.md` changes on push or pull requests
- regenerates the marked mechanical-facts block with `scripts/backlog_facts.py`
- fails closed when the tracked `BACKLOG.md` differs
- regression-tests both workflow triggers and the drift commands

## Stack
- base: `codex/789-backlog-facts` / #950
- head: `codex/789-backlog-facts-ci`
- exact commit: `dfdf68c`

## Evidence
- `python -m unittest discover -s scripts -p "test_backlog_facts.py" -v`: 4 passed
- `python -m unittest discover -s scripts -p "test_*.py"`: 487 passed
- `python scripts/backlog_facts.py` twice + `git diff --exit-code BACKLOG.md`: idempotent, no drift
- `pwsh scripts/verify.ps1`: passed (tree-sitter IXQL 10/10; BAML client in step)
- independent spec and standards reviews: no remaining P0/P1

## Human gate
This PR intentionally remains draft because it changes `.github/workflows/**`. No protected-path review label is self-applied, and no merge is attempted.


spareilleux (commented):
## Cross-Model Review

### Findings

- [CORRECTNESS] `backlog-facts` job runs `python scripts/backlog_facts.py` with no dependency install step shown in this diff — if the script has third-party dependencies, this will fail silently in fresh runners unless they're stdlib-only. Not verifiable from the diff alone; worth confirming.

- [CORRECTNESS] The new job regenerates `BACKLOG.md` in place and then diffs it, but does not restore the file afterward or isolate the checkout. Fine for CI (ephemeral runner) but worth noting the job mutates tracked files as a side effect rather than doing a dry-run comparison — acceptable given the intent, just flagging the pattern.

- [ANTI-SLOP] `test_governance_workflow_enforces_backlog_fact_drift` parses the YAML workflow file via string splitting on literal markers (`"  push:\n"`, `"permissions:"`, etc.) instead of using a YAML parser. This is brittle — any reordering of top-level keys, comment insertion, or whitespace change in the workflow file breaks the test with a confusing `IndexError` rather than a meaningful assertion failure. A `yaml.safe_load` + structural check would be far more robust and less cargo-cult.

- [ANTI-SLOP] The test hardcodes assumptions about section ordering (`permissions:` after both trigger blocks, `workflow_dispatch:` present) that aren't documented as invariants anywhere else. This couples the test tightly to incidental file structure rather than behavior.

- [CORRECTNESS] Minor: the test doesn't verify the two path entries are added under `paths:` specifically (could match a comment or a different list) — string containment checks are permissive enough that the test could pass on accidental placement.

- [LOLLI DETECTION] Named consumer exists (the CI job fails PRs on stale `BACKLOG.md`), and the test asserts the enforcement exists — value is proportional to a small governance surface. Not a lolli.

- [CONSTITUTIONAL] Article 3 (Reversibility): change is additive and easily revertible (remove job + path filter). Article 4 (Proportionality): a 5-minute timeout, single mechanical check — proportionate to the stated problem (stale derived facts). Article 9 (Bounded Autonomy): the job only fails/reports, doesn't auto-commit or auto-merge — stays within bounds.

### Verdict
APPROVE — logic and governance intent are sound; the brittle string-parsing test is a maintainability wart worth fixing but not a blocker.

---

spareilleux:
## Human approval record — 2026-08-03

The repository owner explicitly instructed Codex to apply the ordered merge plan and resolve the human guardrails ("ok applique ca, figure les gardes fous humains aussi").

The Agent Blackbox report was reviewed before applying the override:
- risk `0.70`, caused solely by the protected `.github/workflows/governance-validate.yml` path;
- 2 changed files / 44 changed lines;
- evidence health `7/7` and all functional checks green;
- no implementation defect reported;
- scope is limited to deterministic BACKLOG fact-drift enforcement.

This comment records the explicit human authorization required by `docs/review-independence.md`. The `agent-blackbox-reviewed` label is applied only to this exact reviewed diff.

spareilleux (commented):
## Cross-Model Review

### Findings

- [CORRECTNESS] `scripts/backlog_facts.py` is invoked in the new CI job but not included in this diff — its behavior (idempotency, exit codes, failure modes on malformed BACKLOG.md) is unverifiable from this review alone. Assuming it exists and is tested elsewhere, but flagging the blind spot.

- [CORRECTNESS] The new job trusts `git diff --exit-code` after running the generator script directly against the checked-out branch. If `backlog_facts.py` has any nondeterministic output (timestamps, unordered dict/set iteration, locale-dependent formatting), this job will flake. No evidence of determinism guarantee in the diff.

- [ANTI-SLOP] `test_governance_workflow_enforces_backlog_fact_drift` parses YAML via string `.split()` chains on literal section markers (`"permissions:"`, `"  push:\n"`, `"  pull_request:\n"`, `"  workflow_dispatch:\n"`). This is brittle — any reordering of workflow keys, whitespace change, or added top-level key breaks the test with a confusing `IndexError`/`ValueError` rather than a meaningful assertion failure. A YAML parser (`yaml.safe_load`) would be more robust and only marginally more code.

- [ANTI-SLOP] The test doesn't actually validate correctness of triggers, just presence of substrings in crudely-sliced sections. It gives a false sense of rigor — it will pass even if `BACKLOG.md` were placed inside a comment or a different unrelated job.

- [CORRECTNESS] No handling for the case where `BACKLOG.md` doesn't exist yet (fresh clone edge case) — `git diff --exit-code BACKLOG.md` on a missing file behaves differently than expected; worth confirming the generator creates it if absent, otherwise the job silently no-ops instead of failing loudly.

- [CONSTITUTIONAL / Article 4 - Proportionality] Adding a new CI job, a path filter change in two places, and a meta-test asserting the workflow's own internal structure is a lot of governance surface for keeping a changelog-style doc in sync. The comment justifying it ("status alone would also report unrelated untracked runner files") is reasonable but the mitigation (full regenerate + diff + a test asserting the YAML shape) is heavier than the problem being solved — verges on over-engineering a documentation freshness check.

- [LOLLI DETECTION] The new test doesn't test `backlog_facts.py` logic — it tests that a workflow YAML file contains certain strings. This is a test *about CI configuration*, owned by no clear consumer (it will drift out of sync with the workflow itself and require maintenance in lockstep, doubling the maintenance surface for a single feature). Borderline artifact-without-consumer: it exists to reassure reviewers the wiring is correct, not to catch real regressions in `backlog_facts.py` behavior.

- [CONSTITUTIONAL / Article 3 - Reversibility] The CI gate is a hard failure (`exit 1`) blocking merges if BACKLOG.md drifts. Reasonable given the ADR-0005 precedent cited for baml_client, but worth confirming there's an escape hatch (e.g., admin override / documented bypass) consistent with how the existing baml-drift check is handled, since this doc is lower-stakes than generated client code.

### Verdict
REQUEST_CHANGES — replace the string-slicing YAML test with a proper parser (or drop it in favor of testing `backlog_facts.py` directly), and confirm generator determinism/missing-file behavior before merging.

---

spareilleux (commented):
## Cross-Model Review

### Findings

- [CORRECTNESS] `git diff --exit-code BACKLOG.md` inside `if ! ...; then` — this works, but note that `--exit-code` is redundant with `git diff`'s default behavior when checking exit status; not a bug, just slightly noisy. No functional issue.

- [CORRECTNESS] The new CI job regenerates `BACKLOG.md` via `python scripts/backlog_facts.py` then diffs it. If `backlog_facts.py` itself has a bug that produces nondeterministic output (e.g. unordered dict/set iteration affecting file enumeration order), this job will flake. Not verifiable from the diff alone — worth confirming the script sorts its inputs deterministically.

- [ANTI-SLOP] `test_governance_workflow_enforces_backlog_fact_drift` parses YAML via string `.split()` on literal section markers (`"permissions:"`, `"  push:\n"`, etc.) rather than a YAML parser. This is brittle: any reordering of workflow keys or indentation change silently breaks the test with a confusing `IndexError`/`ValueError` rather than a meaningful assertion failure. A `yaml.safe_load` + structural check would be more robust and less "cargo-cult string surgery."

- [ANTI-SLOP] The test duplicates knowledge of the workflow's structure (job/section names) instead of testing behavior. This is a test that checks "the file contains these substrings in these regions" — it will pass even if the job is broken (e.g., wrong Python version, missing checkout) and will fail on harmless reformatting. Low value per fragility introduced.

- [LOLLI DETECTION] The new `backlog-facts` CI job is a **named, consumed artifact** — it directly enforces that `scripts/backlog_facts.py` output stays synced with `BACKLOG.md`, closing a real drift risk the comment explains clearly (untracked runner files, stale values). This is proportionate: small, single-purpose, self-explanatory. Not a lolli.

- [LOLLI DETECTION] The new test, however, is weaker justification-wise — it doesn't test `backlog_facts.py` logic, it tests that a YAML file contains certain strings in certain regions. This is closer to a low-value governance-weight addition: it exists to say "we tested the CI change" but provides shallow guarantees while adding maintenance burden (breaks on unrelated reformatting).

- [CONSTITUTIONAL: Article 4 - Proportionality] The CI job itself is proportionate — 5 min timeout, single script invocation, clear failure message pointing to remediation. Good.

- [CONSTITUTIONAL: Article 3 - Reversibility] Fully reversible; adds a CI gate, no destructive or irreversible action. Compliant.

- [CONSTITUTIONAL: Article 9 - Bounded Autonomy] The workflow only fails the build and prints a message — it does not auto-commit or auto-fix `BACKLOG.md`. Correctly bounded (fails loud, doesn't act autonomously).

### Verdict
**APPROVE** — CI enforcement is proportionate and reversible; the brittle string-parsing test is a minor anti-slop concern worth a follow-up refactor (use `yaml.safe_load`) but not a blocker.

---

PR #956, opened 2026-08-03
fix(harness): finalize Jules budget reservations

### LINKED ISSUE #937: jules-auto-delegate reserves metered budget and never releases it — arms the moment an approval is committed

spareilleux:
## Summary

- add an idempotent, fail-closed `--abandon-open-provider` terminal path that recovers the sole open provider reservation from the locked cycle ledger
- finalize Jules budget state immediately after `jules-action`, before fallible result reporting, without trusting step outputs
- replace the dangerous job-wide timeout with bounded post-reservation steps and upload explicitly validated ledgers through a SHA-pinned action

## Evidence

- red first: 6 workflow-wiring failures reproduced the missing finalizer and cleanup margin
- `python -m unittest discover -s scripts -p "test_*.py"` — 497 tests passed
- `pwsh scripts/verify.ps1` — IXQL 10/10; BAML client in step
- two independent reviews: no P0/P1 after cancellation-window and timeout fixes

## Risk / human gate

This changes a metered workflow and its budget CLI, so the PR remains draft for human review. It does **not** activate Jules or repair the separate unstable approval identity described on #937; no approval artifact should be committed until that two-phase protocol is fixed.

Fixes #937


spareilleux (commented):
## Cross-Model Review

### Findings

- [CORRECTNESS] `abandon_open_provider`'s "multiple matches → fail closed" path raises `ValueError` uncaught at the CLI boundary — need to confirm (not shown in diff) that this maps to a nonzero exit code rather than an unhandled traceback that could be mistaken for a different failure mode. If the CLI wrapper for `--abandon-open-provider` doesn't explicitly catch and translate this, the workflow's `set -euo pipefail` will fail the step, which is the intended fail-closed behavior — but confirm this is tested, since the diff doesn't show the argparse/CLI dispatch code.

- [CORRECTNESS] Removed job-level `timeout-minutes: 10` in favor of per-step timeouts (5+2+5+2+2+2 = 18 min sum, plus untimed steps like "Validate terminal ledgers" and "Upload artifact"). Net effect is a *longer* possible wall-clock bound than before, and there's no overall backstop if steps individually stay within their limits but checkout/setup steps (untimed) hang. Worth confirming this tradeoff is intentional — the docs justify it ("no job-level timeout that can pre-empt cleanup") but this should be stated explicitly as a deliberate widening of the outer bound, not just a shift.

- [CORRECTNESS] `Record confirmed delegation` and `Report failed delegation` conditions are correctly complementary on `steps.jules.outcome` now (success vs not-success), fixing the prior implicit failure-only gating. Good catch — but verify `steps.jules.outcome` is set even when the step is skipped due to `if: steps.gate.outputs.proceed == 'true'` evaluating false; `outcome` for a skipped step is `'skipped'`, which correctly falls into "not success" and won't spuriously trigger the failure report since that branch also requires `steps.gate.outputs.proceed == 'true'`. This is fine, just fragile — the correctness of both terminal branches depends on precise understanding of skipped-step `outcome` semantics, and there's no test asserting this in the diff.

- [CORRECTNESS] `_abandon_in_cycle` decrements `cycle["active_packets"] -= 1` and `reserved_cost_usd -= estimate` without a floor check (no assertion `active_packets >= 0` or `reserved_cost_usd >= 0` post-decrement). If ledger state is ever corrupted or this is called twice against a stale read, this can silently go negative. Given this is a financial/budget ledger, an invariant assertion here is cheap insurance.

- [ANTI-SLOP] Code duplication between `abandon` and `_abandon_in_cycle` was correctly refactored — this is the *right* kind of change (shared mutation logic extracted, callers differ only in lookup strategy). No complaint here.

- [SECURITY] `google-labs-code/jules-action@v1.0.0` is pinned by tag, not SHA, while `actions/upload-artifact` is correctly pinned to a SHA in the same diff. Inconsistent supply-chain pinning — the Jules action is the one actually receiving `secrets.JULES_API_KEY`, making it the higher-value target for a tag-mutation attack. This should be pinned to a commit SHA like the other action.

- [SECURITY] No issue found with the artifact upload scope — it's explicitly restricted to two named `.octo/*.json` files despite `include-hidden-files: true`, and the PR body itself notes "no other `.octo/` state is published." Correctly scoped.

- [CONSTITUTIONAL: Article 3 / Reversibility] The abandon-not-release design is a good reversibility posture: it never fabricates a verified receipt for a cost that can't be confirmed (`receipt_verified: False`), and pushes the discrepancy into `unreconciled`/`unverified_cost_usd` rather than silently zeroing it out. This is the correct failure-closed behavior for financial state.

- [CONSTITUTIONAL: Article 4 / Proportionality] The finalizer, ledger-validation, and artifact-upload steps add three new steps and ~60 new lines of workflow YAML plus ~70 lines of Python for a genuinely narrow failure window (cancellation between reservation-write and output-publish). This is proportionate *given* the domain is a budget ledger that must not leak reservations — the cost of a stuck "open" reservation blocking future runs is high enough to justify the mechanism. Not a lolli: it has a named consumer (the next scheduled run's preflight, which reads `active_packets`), and the artifact upload has a stated audit purpose, not a speculative one.

- [LOLLI DETECTION] The uploaded artifact (`aiw-budget-ledgers-<run_id>-<run_attempt>`) is explicitly documented as **not** consumed by any later run ("no subsequent workflow downloads the artifact for admission"). This is audit/debug evidence only. That's an honest disclosure rather than a hidden lolli, but it's worth flagging: an artifact with no consumer is inherently a candidate for silent rot (nobody notices if upload starts failing except via `if-no-files-found: error`, which is present — good). Recommend a follow-up issue/TODO for either a retention policy or a defined audit consumer, otherwise this is governance weight (a permanent CI artifact habit) with only diffuse, unclaimed value.

- [ANTI-SLOP] Comments throughout are dense but substantive — they explain *why*, not *what*, and directly address non-obvious race conditions (cancellation timing, skipped-step semantics). This is the opposite of cargo-cult; no complaints.

### Verdict
REQUEST_CHANGES — pin `google-labs-code/jules-action` to a commit SHA (currently the only unpinned secret-consuming action in the diff), and add a floor/invariant check on `active_packets`/`reserved_cost_usd` in `_abandon_in_cycle` before merge.

---

spareilleux:
### Human-gate resolution — AI-first protected-workflow review

- **AI evidence:** Cross-Model Code Review passed and every functional/security check except Agent Blackbox passed. The Gaia-coordinated Claude Code triage independently identified #956 → #959 as a single sequential stack because both touch `.github/workflows/jules-auto-delegate.yml`.
- **Risk interpretation:** Agent Blackbox score 0.70 is caused by the protected workflow path, not by a reported implementation defect. Evidence health is complete.
- **Change scope:** finalize/release Jules budget reservations and cover the consumer with focused tests. This PR does not dispatch Jules, reveal a secret, or spend credits by itself.
- **Human authority:** the repository owner explicitly authorized applying the human safeguards after AI-first review.
- **Decision:** approve `agent-blackbox-reviewed`, promotion from draft, and a fresh update to master. All required checks must pass; no check bypass and no paid workflow dispatch are authorized.

spareilleux (commented):
## Cross-Model Review

### Findings

- [CORRECTNESS] `abandon_open_provider`'s "no reservation" no-op path returns without ever running the ledger-file existence/mkdir logic — fine functionally, but the diff is truncated at the end (`cycle_path.write_text(json.dumps(...` is cut off mid-line with `- ` as the last visible line). If this reflects the actual merged file, it's a syntax error. Needs verification the file is complete and this isn't an artifact of the diff truncation.

- [CORRECTNESS] `if len(matches) != 1: raise ValueError(...)` on multiple matches — this raises inside a `--abandon-open-provider` invocation called with `run: set -euo pipefail`, which will fail the finalizer step. Given the finalizer is the last line of defense before ledger upload, confirm this failure mode is intended to fail the job "closed" per the docs (it is, per the docs, but the workflow step itself has no explicit handling — it just dies noisily, which is consistent with stated intent but worth flagging as a hard dependency between script exit code and workflow correctness).

- [CORRECTNESS] `Record confirmed delegation` condition changed to require `steps.jules.outcome == 'success'`, but `Report failed delegation` requires `steps.jules.outcome != 'success'` AND `(failure() || cancelled())`. If the Jules step is *skipped* (e.g., `gate.outputs.proceed != 'true'` — though both are further gated on `proceed == 'true'` so this shouldn't occur) this is fine, but if Jules step outcome is `'skipped'` for any other reason while `proceed == 'true'` and the job overall didn't fail/cancel, neither the confirm nor the failure-report step fires. Silent gap — narrow but worth a comment justifying why it's unreachable.

- [SECURITY] `abandon-open-provider` reason string is attacker-influenced indirectly (hardcoded here, low risk) but the pattern of truncating to 200 chars (`str(reason)[:200]`) and storing in a JSON ledger is fine — no injection vector since it's not shell-interpolated. No issue, noted only because it's adjacent to genuine risk patterns.

- [CONSTITUTIONAL — Article 3 (Reversibility)] The abandon-on-no-receipt design is explicitly irreversible in the sense that it never calls `--release-job`, permanently pushing the cost into `unreconciled`/`unverified_cost_usd`. The doc is honest about this tradeoff. This is acceptable *if* `unreconciled` entries are consumed by a later reconciliation process — the diff doesn't show one. If no downstream consumer reads `unreconciled`, this is a Lolli pattern (see below).

- [LOLLI DETECTION] The `unreconciled` list and `unverified_cost_usd` field are written but no code in this diff reads or acts on them. Same for the uploaded `aiw-budget-ledgers-*` artifact — it's explicitly documented as "not input to a later run" and nothing downloads/consumes it automatically. This is governance weight (new ledger fields, new artifact upload, new script function) added without a demonstrated consumer. The doc claims audit value, but audit value requires someone/something reading it — that's asserted, not shown.

- [ANTI-SLOP] Naming is generally good (`_abandon_in_cycle`, `abandon_open_provider` are clear), and the refactor extracting shared logic from `abandon()` is legitimate deduplication, not cargo-cult.

- [CORRECTNESS] Removed the job-level `timeout-minutes: 10` in favor of per-step timeouts (5+2+5+2+2+2+2 = 20 min ceiling across new/modified steps, plus checkout/setup steps with no timeout). This is a net *increase* in worst-case job duration with no job-level backstop — if a step the diff didn't touch (e.g., checkout) hangs, there's no outer bound anymore. Should retain a generous job-level `timeout-minutes` as defense in depth alongside the granular ones.

### Verdict
REQUEST_CHANGES — confirm the truncated script tail isn't a real syntax error, add a job-level timeout backstop, and identify or build the consumer for `unreconciled`/uploaded ledger artifacts before merging (Lolli risk).

---

PR #959, opened 2026-08-03
fix(harness): make Jules approvals reachable

### LINKED ISSUE #957: jules-auto-delegate approval identity is unreachable across reruns
## Problem

The committed Jules approval handshake is unreachable.

`jules-auto-delegate.yml` currently derives `job_id` from `GITHUB_RUN_ID`. A rerun keeps the original `GITHUB_SHA`/`GITHUB_REF`, so it cannot see an approval committed after the blocked run. A fresh dispatch sees the commit but gets a new run id, changing both `job_id` and `request_sha256`. No committed approval can therefore be both visible and valid.

The block comment also tells the operator to bind the request SHA-256 without printing the exact SHA.

## Decision

Keep the reviewed Git commit as the human authority boundary, but use an explicit two-phase manual dispatch:

1. Dispatch with `issue_number` and a validated, operator-chosen `approval_id`.
2. If no matching approval exists, fail closed and publish the exact stable `job_id`, provider, and request SHA-256/template.
3. Commit `.octo/aiw-approval.json` through normal review.
4. Start a **fresh** dispatch with the same issue and `approval_id`; the request identity is unchanged and the committed approval is visible.
5. Immediately before the single paid Jules invocation attempt, persist a workflow-authored `jules-approval-attempted` marker. If that write fails, Jules is not invoked. The approval is consumed whether the provider attempt succeeds or fails; every retry requires a new reviewed approval id.

Automatic issue/schedule triggers must not invent paid authority. They may route the issue to the manual two-phase path but must not reserve or invoke Jules without an explicit approval id.

The paid credential is isolated as `JULES_PAID_API_KEY` in the `jules-paid-default-branch` GitHub environment, whose deployment policy accepts protected branches only. The repository-wide `JULES_API_KEY` has been removed. The workflow also pins checkout to the repository default branch and fails closed on any other event ref.

## Acceptance

- [ ] `job_id` and request SHA are stable across fresh workflow runs using the same issue + approval id.
- [ ] the blocked path prints an exact approval template, not prose requiring an unknown hash.
- [ ] a fresh dispatch can see and validate the committed approval.
- [ ] approval ids are strictly validated before inclusion in job ids/comments.
- [ ] an approval id is durably consumed immediately before its one paid invocation attempt and cannot be replayed, even with `force` or failed result reporting.
- [ ] GitHub comment-read failures fail closed rather than looking like marker absence.
- [ ] an unreviewed workflow ref cannot receive the paid Jules credential.
- [ ] a different approval id is required for every retry or intentional re-delegation.
- [ ] automatic triggers never synthesize paid approval.
- [ ] tests cover phase 1, phase 2 identity, replay, malformed ids, ref isolation, comment API failure, and force behavior.
- [ ] no auto-merge and no approval artifact is committed as part of the implementation PR.

Related: #937, #956, #789, #471

spareilleux:
## Summary

- replace the unreachable run-id-bound Jules approval with a stable two-phase `issue + approval_id` handshake
- bind the reviewed approval artifact to that exact id and request SHA, while allowing phase one to replace an older one-shot artifact
- consume the one-shot authority durably before the single paid invocation and fail closed on GitHub comment API ambiguity
- invoke the Jules create-session REST endpoint directly, without POST retry, and require a valid `sessions/<id>` response before writing `jules-delegated`
- restrict the paid job to `workflow_dispatch`, the repository default branch, and the protected `jules-paid-default-branch` environment
- document and test replay, ref, cancellation, finalization, and #925 ordering invariants

## Stack

This draft is stacked on **GuitarAlchemist/Demerzel#956** (`codex/937-finalize-jules-budget`). Merge #956 first, then retarget this PR to `master` if GitHub does not do so automatically.

## Control-plane migration already applied

- environment `jules-paid-default-branch` created with `protected_branches=true`
- only protected branch currently present: `master`
- environment secret `JULES_PAID_API_KEY` installed
- old repository-wide `JULES_API_KEY` removed

Until #956 and this PR merge, Jules API delegation intentionally remains fail-closed.

## Evidence

- `python -m unittest discover -s scripts -p "test_*.py"` — 510/510 passed
- `pwsh scripts/verify.ps1` — passed
- IXQL — 10/10 passed
- BAML client — in step
- two independent read-only reviews — no P0/P1

No auto-merge, approval artifact, provider call, or human-review override is included.

Fixes #957
Related: #937, #956, #789, #925


spareilleux (commented):
## Cross-Model Review

### Findings

- [CORRECTNESS] `has_approval_marker` / `has_delegation_marker` rely on `bot_comments()` grep matching, but `gh api --paginate` failure modes (rate limit, transient 5xx) are conflated with "no comments found" only via the `return 2` path — verify that *every* non-zero `gh api` exit actually propagates through the `if ! comments=...` check and isn't swallowed by `set -euo pipefail` semantics inside a function called in an `||` context (functions run with `set -e` disabled in the calling `if`/`||` context in bash, so this should work, but it's worth an explicit test since this is a fail-closed security boundary).

- [CORRECTNESS] The truncated diff cuts off mid-heredoc (`comment_bo`) — cannot verify the final `gh issue comment` call or exit path for the budget-blocked branch. Review is incomplete on this critical path; the PR should not merge until this section is visible and reviewed in full.

- [CORRECTNESS] `approval_id` regex `^[A-Za-z0-9][A-Za-z0-9._-]{0,63}$` allows a single character total, and permits `.` which could enable path-adjacent tricks if `approval_id` is ever used to construct file paths (currently it's only used in job_id strings and JSON, so low risk today — but flag if reused elsewhere later).

- [CORRECTNESS] Stale approval artifact handling: `mv "$approval_artifact" "$RUNNER_TEMP/stale-aiw-approval.json"` — if this mv fails (permissions, RUNNER_TEMP unset in some runner contexts), the script has `set -e` so it will abort, which is fail-closed and correct, but there's no explicit error message, making failures harder to diagnose.

- [SECURITY] Good: moving to a protected `jules-paid-default-branch` environment plus `if: github.event_name == 'workflow_dispatch'` at the job level is a strong control — this correctly prevents `issues:labeled`/`schedule` triggers from ever reaching a job that has environment secrets in scope. This is the right pattern (job-level gating, not just step-level).

- [SECURITY] `ref: ${{ github.event.repository.default_branch }}` combined with the `REF_NAME != DEFAULT_BRANCH` check is redundant but not harmful — actually this is a meaningful defense-in-depth: checkout pins ref explicitly, and the runtime check guards against workflow-level ref confusion. Acceptable belt-and-suspenders, not slop.

- [SECURITY] Approval artifact trust: `.octo/aiw-approval.json` is read from the checked-out ref content, meaning anyone who can land a commit on default branch (or a PR that gets merged) controls `approver` and `approved_at` fields, which are explicitly templated as `REPLACE_WITH_*` placeholders — good, this signals human-in-the-loop intent, but there's no verification anywhere in view that `approver` is checked against a real reviewer identity (e.g., matching `github.actor` or a CODEOWNERS-style check). If that verification exists in the cut-off portion, fine; if not, the approval gate is honor-system enforced by commit review alone, which may be acceptable given branch protection but should be explicit in comments.

- [ANTI-SLOP] Variable naming is clear and intention-revealing (`approval_marker_rc`, `committed_approval_id`, `delegation_marker_rc`). No complaints here — this is well above average for shell-in-YAML.

- [CONSTITUTIONAL: Article 9 / Bounded Autonomy] Strong alignment: this diff *tightens* autonomy bounds rather than loosening them — removing label/schedule-triggered paid dispatch, requiring an explicit `approval_id` typed twice (phase 1/phase 2) by a human operator, and failing closed (`exit 1`, `rc == 2` paths) on any ambiguity in delegation/approval state. This is a textbook proportional response to a prior lolli risk (schedule-triggered spend).

- [CONSTITUTIONAL: Article 3 / Reversibility] The `force` + fresh `approval_id` re-delegation path is reversible and auditable — old approval markers remain in comments, new ones are additive, nothing is deleted. Good.

- [LOLLI CHECK] Not a lolli — this modifies an existing governed workflow to close a real gap (paid API triggerable by non-human events) and every new artifact (`approval_id` input, `.octo/aiw-approval.json`, approval marker comments) has a named consumer in the same file (the gate logic itself, and the human operator who must supply matching ids across two dispatch calls). Governance weight added is proportional to the risk removed (unattended paid spend).

### Verdict

REQUEST_CHANGES — diff is truncated mid-script (missing `comment_bo...` and beyond), so the final gate/exit logic and comment content cannot be fully verified before approval.

---

spareilleux:
CI terminal state: all functional checks are green (governance validate, 510 Python tests, BAML drift, manifest/beliefs, cross-model detect-and-review). Agent Blackbox \
isk-report\ is the sole red check by policy: risk 0.85 because \.github/workflows/jules-auto-delegate.yml\ is an autonomy-blocked path and the diff exceeds 300 lines; evidence health is 7/7 and it reports no implementation defect. This draft therefore remains at the intended human-review gate. No override label has been applied.

spareilleux:
### Human-gate resolution — AI-first protected-workflow review

- **AI evidence:** Cross-Model Code Review passed; all functional checks passed on the stacked branch. The Gaia-coordinated Claude Code triage independently confirmed that #956 must land before #959 because both change `.github/workflows/jules-auto-delegate.yml`.
- **Risk interpretation:** Agent Blackbox score 0.85 comes from the protected workflow path plus the change size, not from a reported implementation defect. Evidence health is complete.
- **Change scope:** two-phase stable Jules approval, protected environment, and one-shot approval consumption with focused ordering/handshake tests. This PR does not dispatch a paid Jules job, reveal a secret, or spend credits by itself.
- **Human authority:** the repository owner explicitly authorized applying the human safeguards after AI-first review.
- **Decision:** approve `agent-blackbox-reviewed`, retargeting to master after #956, and promotion from draft. Fresh required checks must pass; no check bypass and no paid workflow dispatch are authorized.

spareilleux (commented):
## Cross-Model Review

### Findings

- [CORRECTNESS] `jq -er '.request_sha256' .octo/aiw-budget-ledger.json` (and the `reasons` line above it) now hard-fails with `set -e` if the ledger file is missing or malformed, whereas the old code had a `2>/dev/null || echo "budget policy"` fallback. If `aiw_budget_gate.py` returns nonzero without writing a ledger (e.g., crash before write), this branch will abort with an unhandled error instead of degrading to a generic message — a regression in robustness for an already-failure-prone path.

- [CORRECTNESS] The diff is truncated mid-line (`comment_bo`), so the full behavior of the approval-template comment path can't be verified. Flagging as incomplete review coverage — this section needs to be seen in full before sign-off.

- [CORRECTNESS] `has_delegation_marker`/`has_approval_marker` now search only `github-actions[bot]` comments via `bot_comments`, whereas the old code scanned all issue comments via `gh issue view --json comments`. This is actually a security improvement (prevents a non-bot commenter from spoofing the marker string to block/unblock delegation) but it's an undocumented behavior change worth calling out explicitly in the PR description/comments — it changes what counts as "already delegated."

- [CORRECTNESS] Return-code convention (0 = found, 1 = not found, 2 = error) is implicit and relies on `grep`'s exit codes plus a manual `return 2` — fragile if `bot_comments` fails for a reason other than API error (e.g., empty result set causing `grep` to return 1, which is correctly distinguished, but this depends on `gh api --paginate` never producing exit code 2 semantics itself). Low risk but worth a comment explaining the code contract.

- [SECURITY] Good: `approval_id` is validated with a strict regex before being interpolated into shell/JSON/markdown contexts, preventing injection via workflow_dispatch input. This is a correct and necessary control given the input is user-supplied.

- [SECURITY] Good: moving to `environment: jules-paid-default-branch` with a `ref` pin to the default branch and gating on `workflow_dispatch` only is a strong reduction in attack surface — labels/schedule can no longer reach paid credentials. This directly addresses a real privilege-escalation risk in the prior design (bot-applied labels triggering paid API calls).

- [CONSTITUTIONAL COMPLIANCE — Article 9 Bounded Autonomy] The two-phase approval (approval_id must match between dispatch and a committed `.octo/aiw-approval.json`) correctly binds paid execution to a specific human-reviewed artifact rather than a generic re-run. This is a meaningful bounded-autonomy improvement over the previous single `JULES_API_KEY` presence check.

- [CONSTITUTIONAL COMPLIANCE — Article 3 Reversibility] Renaming the stale approval artifact to `$RUNNER_TEMP/stale-aiw-approval.json` instead of deleting it preserves audit evidence — correct handling for reversibility/auditability.

- [ANTI-SLOP] Naming is precise throughout (`approval_marker_rc`, `delegation_marker_rc`, `committed_approval_id`) — no vague variables. Comments explain *why*, not just *what*. This is well above the median bar for workflow YAML.

- [LOLLI DETECTION] No governance artifacts created without consumers here — the `.octo/aiw-approval.json` template is consumed by the same workflow's gate check, and the approval_id marker is consumed by both `has_approval_marker` and the budget-gate comment flow. Weight added (extra validation steps, environment gating) is proportional to the risk being mitigated (paid, irreversible external API spend).

### Verdict
REQUEST_CHANGES — diff is truncated (incomplete review possible) and the budget-ledger fallback removal (`jq -er` without `2>/dev/null` fallback) introduces a new hard-failure mode that should be confirmed intentional.

---

spareilleux (commented):
## Cross-Model Review

### Findings

- [CORRECTNESS] `bot_comments()` filters strictly on `user.login == "github-actions[bot]"`. If the delegation/approval-marker comment is ever posted under a different bot identity (e.g., GITHUB_TOKEN posting as a different app, or a future migration to a GitHub App identity), `has_delegation_marker`/`has_approval_marker` will silently return "not found" (rc=1) rather than failing closed. This is a fail-open path disguised as fail-closed logic — worth an explicit comment/test pinning the identity assumption.

- [CORRECTNESS] The approval-consumption check only searches issue comments for the marker. If comment history is edited, deleted, or the issue is transferred, the same `approval_id` could be replayed. There's no independent ledger (e.g., a file under `.octo/` or `governance/state/`) recording consumed approval ids, so the anti-replay guarantee is only as strong as GitHub comment integrity — reasonable for now, but should be documented as a known limitation, not implied as strong.

- [CORRECTNESS] Diff is truncated mid-heredoc (`comment_bo` at the end) — cannot verify the final `gh issue comment` call and downstream logic (e.g., whether `exit 1` still occurs after posting the approval template, and whether the stale-approval-artifact restoration/cleanup is handled). Flag for re-review once the full diff is available.

- [CORRECTNESS] Regex `^[A-Za-z0-9][A-Za-z0-9._-]{0,63}$` for `approval_id` is fine, but this value is later interpolated directly into `job_id` and shell variables without additional quoting review in all call sites — low risk given the regex, but confirm no path uses `approval_id` in a `jq` filter string or similar without `--arg`.

- [SECURITY] Good: moving from ambient `JULES_API_KEY` to an environment-scoped `JULES_PAID_API_KEY` behind `environment: jules-paid-default-branch`, restricting to `workflow_dispatch`, and pinning `ref` to the default branch are all strong, deliberate reductions of blast radius. No secrets are echoed or logged.

- [SECURITY] `mv "$approval_artifact" "$RUNNER_TEMP/stale-aiw-approval.json"` — moving a stale approval artifact into runner temp preserves it as "run evidence" per the comment, but runner temp is ephemeral and not uploaded as an artifact anywhere in this diff (as far as shown). If the intent is auditability, this data is lost when the job ends. Either upload it as a workflow artifact or drop the "preserve as evidence" claim.

- [ANTI-SLOP] Naming is precise and intention-revealing throughout (`approval_marker_rc`, `committed_approval_id`, `bot_comments`) — no vague identifiers. This is a well-annotated diff with reasoning comments at each gate, which is appropriate given the stakes (paid API invocation).

- [LOLLI DETECTION] No lolli pattern here — every new artifact (`approval_id` input, `.octo/aiw-approval.json`, approval marker comments) has a clear named consumer: the budget gate reads the approval file, the workflow reads the marker for replay protection, and the human approver is the explicit actor closing the loop. This adds real governance weight but ties it to genuine risk (uncontrolled spend) — proportional.

- [CONSTITUTIONAL — Article 4 Proportionality] Strong compliance: the change adds friction (explicit two-phase approval, approval_id binding, environment gating) specifically for a real risk (metered paid API spend), not for low-stakes operations. This is proportional governance, not governance-for-its-own-sake.

- [CONSTITUTIONAL — Article 3 Reversibility] The `force` flag combined with mandatory fresh `approval_id` correctly prevents silent re-delegation while still allowing an explicit, reviewed override — good reversibility posture. Minor gap: no logged trail of *why* force was used beyond the new approval_id itself.

- [CONSTITUTIONAL — Article 9 Bounded Autonomy] Correctly bounds autonomy: `if: github.event_name == 'workflow_dispatch'` hard-stops the job for label/schedule triggers before any credential exposure, and the code explicitly states these paths "can never receive credentials or synthesize paid authority." This is the right mechanism (job-level `if`, not just in-script checks) since it prevents the job from even starting.

### Verdict
REQUEST_CHANGES — diff is truncated (incomplete review of final approval-comment/exit logic) and the bot-identity assumption in `bot_comments()` is a fail-open risk that should be hardened or explicitly documented before merge.

---

PR #960, opened 2026-08-03
fix(harness): declare cycle driver execution policy

### LINKED ISSUE #776: run_afk_cycle.py and run_ml_feedback_cycle.py are scheduled by no workflow (producer-without-consumer)

spareilleux:
## Summary

- declare `run_afk_cycle.py` as manual-or-external, with no GitHub Actions schedule or automatic freshness guard because provider execution can consume metered budget
- declare the mutating ML-feedback cycle as externally scheduled while keeping `ml-governance-schedule.yml` a read-only freshness tripwire and strict dry-run smoke
- add focused contract tests for both declarations and the named tripwire configuration

## Verification

- `python -m unittest discover -s scripts -p "test_*.py"` — 485 tests passed
- `pwsh scripts/verify.ps1` in an LF checkout — IXQL 10/10; BAML generated client in step
- two independent read-only reviews — no P0/P1/P2 findings

## Risk

Documentation and contract-test only. No workflow, cadence, provider invocation, budget, or runtime behavior changes.

Fixes #776

### DIFF
--- a/scripts/run_afk_cycle.py
+++ b/scripts/run_afk_cycle.py
@@ -2,6 +2,12 @@
 """
 Demerzel run_afk_cycle — the AFK implement-lane governor.
 
+Execution policy: manual-or-external.
+No GitHub Actions workflow schedules this driver. An operator or external harness
+owns cadence and invokes one bounded cycle after issue authorization, HALT, and
+budget preflight. Because provider runs can spend metered budget, no automatic
+freshness guard or self-starting schedule is wanted.
+
 Reads the `agent-implement` GitHub issue queue, honors HALT, classifies risk per
 policies/autonomous-loop-policy.yaml, and for each eligible (non-critical) issue
 invokes the agent-agnostic sandcastle harness (../afk-harness) which runs headless
--- a/scripts/run_ml_feedback_cycle.py
+++ b/scripts/run_ml_feedback_cycle.py
@@ -3,6 +3,11 @@
 Demerzel run_ml_feedback_cycle — the orchestrator that turns the four hand-run
 ML-feedback legs into ONE unattended, bounded self-improvement cycle.
 
+Execution policy: externally-scheduled.
+An operator-managed cron or Task Scheduler owns the mutating cycle's cadence.
+GitHub Actions `ml-governance-schedule.yml` is a read-only freshness tripwire and
+strict dry-run smoke; it never runs the mutating cycle.
+
 It runs, in dependency order, a single pass of pipelines/ml-feedback-loop.ixql
 (standing in for the absent IxQL executor):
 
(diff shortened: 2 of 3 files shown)


spareilleux (commented):
## Cross-Model Review

### Findings

- [CORRECTNESS] `_module_docstring` uses `ast.get_docstring`, which dedents/normalizes whitespace — the assertion `self.assertIn("no automatic\nfreshness guard", docstring)` is fragile and depends on exact line-wrap position in the source docstring. Any reflow of the comment (even for line length) silently breaks this test without indicating a policy violation, producing a false-negative-prone brittle test rather than a semantic check.

- [CORRECTNESS] `workflow.get("on", workflow.get(True, {}))` — this works around YAML 1.1's `on:` being parsed as boolean `True` by PyYAML, but it's a fragile hack. If the workflow file is later minorly reformatted (e.g., quoted `"on":`), this silently returns `{}` and `self.assertIn("schedule", triggers)` fails opaquely with no explanation of *why*. A comment explaining the YAML gotcha would prevent future confusion.

- [CORRECTNESS] No test asserts that `run_ml_feedback_cycle.py`'s actual *behavior* (not just docstring text) refuses to self-schedule or lacks a freshness gate — this is a documentation/contract test, not a behavioral one. The test enforces that words exist in a comment, not that the code actually behaves per the stated policy. Mislabeling risk: a reviewer might believe this provides behavioral guarantees it doesn't.

- [ANTI-SLOP] Test coupling to exact CLI argument order via `shlex.split` + list equality is over-specified. `--dry-run --strict --repos-root _siblings` in a different order would fail the test despite being behaviorally identical. Prefer set-based or substring checks for flags.

- [ANTI-SLOP] `_module_docstring` duplicated inline rather than as a small reusable fixture is fine at this scale, but the function conflates "get docstring" with implicit path resolution assumptions (`ROOT / relative_path`) — minor, not blocking.

- [LOLLI DETECTION] This diff is documentation-only (docstring clarifications) plus a test that pins those docstrings and one workflow's shape. It has a named consumer (prevents drift between the "who schedules this" policy and actual code/workflow reality) and low governance weight — it doesn't introduce new autonomous capability, only guards existing stated intent. Reasonable, proportional artifact.

- [CONSTITUTIONAL COMPLIANCE] Article 4 (Proportionality) / Article 9 (Bounded Autonomy): the diff reinforces rather than expands autonomy — explicitly disclaiming self-scheduling and confirming the workflow is read-only/dry-run. This is the correct direction of change under Article 9.

- [CONSTITUTIONAL COMPLIANCE] Article 3 (Reversibility): no state-mutating logic changed; purely comments + a test. No reversibility concerns.

### Verdict
REQUEST_CHANGES — brittle docstring/whitespace-exact assertions and the YAML `on:`-key hack should be made resilient (semantic checks over literal string/order matching) before merge; substance and intent are sound.

---

